PR #13, opened 2016-03-24
Use flexbridge for send/receive (v2)

ermshiperete:
The two working `SynchronizeActionTests` pass with this change.

This change requires a Chorus change (https://github.com/sillsdev/chorus/pull/85) and a FlexBridge change (https://github.com/sillsdev/flexbridge/pull/89).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/13)

Issue #1, opened 2015-10-22
Windows DLL ole32.dll not found on Mono


rmunn:
When I try to run the LfMerge skeleton as it currently stands (as of commit 17269b8), I get an exception in `FwAccess.cs` when it tries to call `FdoCache.CreateCacheFromExistingData`. Apparently one of the methods deep inside FDO still has a dependency on something COM-related, because `SIL.FieldWorks.FDO.IOC.FdoServiceLocatorFactory.CreateServiceLocator ()` is trying to construct a  `SIL.FieldWorks.FDO.IOC.StructureMapServiceLocator`object, which in turn is calling `SIL.FieldWorks.Common.COMInterfaces.LgIcuCharPropEngineClass.Create ()`, which in turn is calling `System.Runtime.Remoting.Activation.ActivationServices.CreateProxyForType (type={SIL.FieldWorks.Common.COMInterfaces._LgIcuCharPropEngineClass})`. And that, after a few more calls, fails because Mono can't find the Windows DLL `ole32.dll`.

Stack trace follows:

```
System.DllNotFoundException: ole32.dll
  at at (wrapper managed-to-native) System.__ComObject:CoCreateInstance (System.Guid,intptr,uint,System.Guid,intptr&)
  at System.__ComObject.Initialize (System.Type t) [0x00082] in <filename unknown>:0
  at at (wrapper remoting-invoke-with-check) System.__ComObject:Initialize (System.Type)
  at Mono.Interop.ComInteropProxy.CreateProxy (System.Type t) [0x00007] in <filename unknown>:0
  at System.Runtime.Remoting.RemotingServices.CreateClientProxyForComInterop (System.Type type) [0x00000] in <filename unknown>:0
  at System.Runtime.Remoting.Activation.ActivationServices.CreateProxyForType (System.Type type) [0x00047] in <filename unknown>:0
  at at (wrapper managed-to-native) object:__icall_wrapper_mono_object_new_specific (intptr)
  at SIL.FieldWorks.Common.COMInterfaces.LgIcuCharPropEngineClass.Create () [0x00001] in <filename unknown>:0
  at SIL.FieldWorks.FDO.IOC.StructureMapServiceLocator..ctor (StructureMap.Container container) [0x00000] in <filename unknown>:0
  at SIL.FieldWorks.FDO.IOC.FdoServiceLocatorFactory.CreateServiceLocator () [0x00448] in <filename unknown>:0
[...]

PR #15, opened 2016-04-20
Use flexbridge for send/receive (v3)

ermshiperete:
This PR depends on [flexbridge PR #92](https://github.com/sillsdev/flexbridge/pull/92) and supersedes #13.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/15)


regnrand:
1. Nothing in the LFMerge repo or any of its tests can reference, or use, anything in any FB assembly, except indirectly via LFBridge. Any prior changes that made internal classes/methods public and/or not sealed, must be reverted.

2. LfMerge should not reference or use anything an any Chorus assembly. That is the point of using FB.

3. You will be better off if you can figure out how to live without Autofac, especially, since it is newer than the one Chorus uses and both of them appear to want to be in the same "assembly soup" folder.

Doing #1 & #2 will then have your unit test effort focus on LfMerge behaviors, and not the behaviors of other code, such as Chorus or FB.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge/LfMerge.csproj, line 42 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-KFtvfntYaRQ9jOlMLh0:-KFtvfntYaRQ9jOlMLh1:-1502706592) ([raw file](https://github.com/sillsdev/lfmerge/blob/af2d00535a9e40323bc6c496fea5d157f609e6b4/src/LfMerge/LfMerge.csproj#L42)):*
Will Autofac 3.5.2 cause problems with the version in Chorus (2.6.3.862)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-:-KGJ0v95l9ghCCxaCghz:-296218358)*

ermshiperete:
I agree that it would be better to move the clone action into FB. That would take care of the first two bullets and get rid of direct dependencies on Chorus (I think) and other FB assemblies. But that should be a separate task and not be included in this PR.

Regarding 3, so far we didn't have problems with Autofac versions. I also send an email to get some clarification why you prefer MEF over Autofac.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge/LfMerge.csproj, line 42 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-KFtvfntYaRQ9jOlMLh0:-KGMlNtGdBH2njmxHlwz:-671460759) ([raw file](https://github.com/sillsdev/lfmerge/blob/af2d00535a9e40323bc6c496fea5d157f609e6b4/src/LfMerge/LfMerge.csproj#L42)):*
So far we didn't run into problems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-:-KGMjPfTaGT6KT0KZ-HP:-1255501660)*

regnrand:
Could I do at least #1 in a prerequisite PR? LfMerge may, or may not, need to use Chorus directly in its handling of chorus notes.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-:-KGNHjnujb3DEpPkmVff:2031441941)*

ermshiperete:
Sure.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-:-KGNWmjEVpfsSrxHpQnJ:-485110604)*

regnrand:
Thanks. I'll get to it today.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/15#-:-KGNgeM9kMFx0E8jBwR-:-1808321207)*

PR #17, opened 2016-04-28
My attempt to remove dependencies on Chorus & Flex Bridge

regnrand:
I didn't run the tests.

I left 'breadcrumbs' here and there with thigns that
still need doing, and what needs to take palce in new calls
into LFBridge (in FB). Those breadcrumbs are identified with:

#if GETTING_RID_OF_FB_AND_CHORUS
#endif

and

// GETTING_RID_OF_FB_AND_CHORUS TODO:

comments.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/17)


ermshiperete:
Reviewed 12 of 12 files at r1.

---

*[src/LfMerge/Program.cs, line 35 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS2g_A61xBumsq0G0d:-KGS2g_A61xBumsq0G0e:-2033750796) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge/Program.cs#L35)):*
The reason I created `UpdateBranchHelper` (and derived classes) was that there was some code that was duplicated for both Lift and FLEx.

I agree that those classes shouldn't be public. To match the philosophy of FLExBridge they should be refactored to use MEF instead of deriving from a superclass, but I still see benefits from moving that duplicated code into one place.

---

*[src/LfMerge.Tests/Actions/ProgramTests.cs, line 36 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS5TXNVC_8cNwaLrt_:-KGS5TXNVC_8cNwaLrta:-1789984126) ([raw file](https://github.com/sillsdev/lfmerge/blob/0b2ef053b7b4059aea384aa50eb4c11aaaea1576/src/LfMerge.Tests/Actions/ProgramTests.cs#L36)):*
I think we still want this test because it also tests that the state gets set properly if cloning fails. Needs to be implemented differently though.

---

*[src/LfMerge.Tests/Actions/SynchronizeActionTests.cs, line 27 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS6Qgho9V2AK-ckGpQ:-KGS6Qgi2dAZIawMX_rf:1727062790) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge.Tests/Actions/SynchronizeActionTests.cs#L27)):*
These tests are only partially testing FB code. I guess these are really more like acceptance tests rather than unit tests, but I think we still need similar tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/17)*

regnrand:

---

*[src/LfMerge/Program.cs, line 35 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS2g_A61xBumsq0G0d:-KGSTO4qouWc8bqecqQL:-485698205) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge/Program.cs#L35)):*
Agreed on not public and heading towards less duplicated code. I'm less excited about classes having subclasses. I need more coffee to know if MEF is needed. ;-)

---

*[src/LfMerge.Tests/Actions/ProgramTests.cs, line 36 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS5TXNVC_8cNwaLrt_:-KGTD7mKQSdLz_6eU7KU:247072423) ([raw file](https://github.com/sillsdev/lfmerge/blob/0b2ef053b7b4059aea384aa50eb4c11aaaea1576/src/LfMerge.Tests/Actions/ProgramTests.cs#L36)):*
It seems you can still skip it the code that runs when the user first enters the LD login credentials then checks with LD to see if they are valid. If not, then there is no need to catch that Chorus.VcsDrivers.Mercurial.RepositoryAuthorizationException exception, and thus, no need to set the set to HOLD on a clone attempt, at least for login in issues.

I had understood that code already asks LD for projects the user can work with, so if they then select one, how can the login fail so soon afterwards? Why not have LfMerge do its clone right away, rather than have always check to see if one is needed?

---

*[src/LfMerge.Tests/Actions/SynchronizeActionTests.cs, line 27 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS6Qgho9V2AK-ckGpQ:-KGSTm148jNXO1JPUhfv:1294612739) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge.Tests/Actions/SynchronizeActionTests.cs#L27)):*
Can you revise SynchronizeAction to have it call FB in the real world, but to call some do-nothing code in tests? Then your pre-test code can set up the context properly for before and after the FB call and you can pretend you really called FB? Chorus takes forever to run its tests, because it does real Hg clones, commits, merges, etc. But then Chorus needs to test for that, since those are its "units".

I know the file splitter/joiner code only has some superficial unit tests. The real tests I run any time any of the file splitter or joiner code is modified in *any* way are done using the separate "FwdataNester" program I wrote. It can do lots of things, but its main reason to exist is to split up the fwdata file and then rebuild it. I then do a diff on the original fwdata file and the rebuilt one.  I better have a really good reason for any differences between the two, or the splitter/joiner change candidates may have broken things. I have 45 real and test FW test data sets I use for this manual testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/17)*

megahirt:
*[src/LfMerge.Tests/Actions/ProgramTests.cs, line 36 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS5TXNVC_8cNwaLrt_:-KGUhyQi7V_k33FUgxPY:971981077) ([raw file](https://github.com/sillsdev/lfmerge/blob/0b2ef053b7b4059aea384aa50eb4c11aaaea1576/src/LfMerge.Tests/Actions/ProgramTests.cs#L36)):*
As it turns out, we will not actually be doing LD authentication (username/password) at all.  We will have a trusted connection between LF and LD that requres no authentication.  I think LFMerge should handle the case where it's asked to clone a non-existent project, and this test should ensure that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/17)*

megahirt:
*[src/LfMerge.Tests/Actions/SynchronizeActionTests.cs, line 27 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS6Qgho9V2AK-ckGpQ:-KGUiyT-SXyUfhHu4ZOq:1264704124) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge.Tests/Actions/SynchronizeActionTests.cs#L27)):*
We still need these tests, but I agree that if there is a way to mock the FB actions in the tests, we should do that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/17)*

megahirt:
*[src/LfMerge.Tests/Actions/SynchronizeActionTests.cs, line 27 \[r1\]](https://reviewable.io:443/reviews/sillsdev/lfmerge/17#-KGS6Qgho9V2AK-ckGpQ:-KGUjAIN0ggq4xhyqGtj:1083775565) ([raw file](https://github.com/sillsdev/lfmerge/blob/8782384d6835c492f4e4a9bcb12153d21ff4f543/src/LfMerge.Tests/Actions/SynchronizeActionTests.cs#L27)):*
But I'll say that that is all future work.  Right now, let's move forward with what we've got because they are working, and work to refine them as we write more tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/17)*

PR #23, opened 2016-05-31
Fix Mercurial in package

ermshiperete:
When building the source package all *.so files get removed. This
breaks Mercurial. So we now keep the Mercurial*.zip file and remove
the extracted files.

NOTE: this is a partial fix. It will need another change in `rules` or `LfMerge.proj` to extract the zip file.


ermshiperete:
Robin fixed it differently

PR #33, opened 2017-01-20
Redo model migration with proper line endings

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/33)


rmunn:
We seem to have worked in parallel on these changes; while you were creating this PR, I pushed commit d922a41 to master that makes the unit tests pass (at least on my machine). And this PR does *not* pass the unit tests on my machine: the three `Success_*` tests in `SynchronizeActionBrideIntegrationTests` fail for me, probably because of the line endings in the `MorphAndSynData.morphdata` file. (The `<Uni>` line apparently should have a LF ending, not a CRLF, for reasons I don't yet fully understand).

There are some parts of this PR that aren't in my version, and that should probably be cherry-picked into master (such as commit 5210f6f). But I don't think we should merge this PR into master just now until we figure out why it doesn't pass tests on my machine and master does pass the same tests.

rmunn:
The current `master` branch also passes unit tests on Jenkins:

https://jenkins.lsdev.sil.org:45192/view/LfMerge/view/All/job/LfMerge-Linux-any-master-release/340/consoleFull

So I think this PR has become unnecessary, though some parts of it should probably be cherry-picked.

rmunn:
Reviewed 610 of 610 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[data/testlangproj/Linguistics/MorphologyAndSyntax/MorphAndSynData.morphdata, line 568 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/33#-Kb8X5xTPOLTSbis_4aM:-Kb8X5xTPOLTSbis_4aN:bs77chm) ([raw file](https://github.com/sillsdev/lfmerge/blob/5210f6ff8b272a3301dca4bb88c008fd8f2e4a3c/data/testlangproj/Linguistics/MorphologyAndSyntax/MorphAndSynData.morphdata#L568)):*
> ```
> 		</CompoundRules>
> 		<ParserParameters>
> 			<Uni>
> ```

In my testing, I discovered that having a CRLF at the end of the `<Uni>` line here caused unit test failures. The only thing that worked was having the `<Uni>` line, and the `&lt;ParserParameters` (etc.) line that follows it, terminated by an LF without CR.

---

*[src/LfMerge.Core.Tests/TestDoubles.cs, line 310 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/33#-Kb8WhsQvaGdJ6Rkvrlt:-Kb8WhsQvaGdJ6Rkvrlu:b-4nheb3) ([raw file](https://github.com/sillsdev/lfmerge/blob/5210f6ff8b272a3301dca4bb88c008fd8f2e4a3c/src/LfMerge.Core.Tests/TestDoubles.cs#L310)):*
> ```Smalltalk
> 				InputSystems = new Dictionary<string, LfInputSystemRecord>() {
> 					{"en", new LfInputSystemRecord {
> 							Abbreviation = "Eng",
> ```

This is good, and should probably be cherry-picked even if we don't merge this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/33)*

ermshiperete:
I saw your changes but decided to continue anyway because it'll fix the root cause of the problem so that next time a model change happens we don't run into this problem again. But since we're still getting failing tests everywhere except on my machine it looks like I need to do some more work.

ermshiperete:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[data/testlangproj/Linguistics/MorphologyAndSyntax/MorphAndSynData.morphdata, line 568 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/33#-Kb8X5xTPOLTSbis_4aM:-Kb9OXSqdRQRouvbrJ34:bs83b49) ([raw file](https://github.com/sillsdev/lfmerge/blob/5210f6ff8b272a3301dca4bb88c008fd8f2e4a3c/data/testlangproj/Linguistics/MorphologyAndSyntax/MorphAndSynData.morphdata#L568)):*

In my testing, I discovered that having a CRLF at the end of the `<Uni>` line here caused unit test failures. The only thing that worked was having the `<Uni>` line, and the `&lt;ParserParameters` (etc.) line that follows it, terminated by an LF without CR.
</blockquote></details>

Yes, the problem is that HG and Git don't agree on the line-ending normalization. I apparently failed to fix it properly last summer (when I tried it in Git), so this PR is another attempt in hg.

---

*[src/LfMerge.Core.Tests/TestDoubles.cs, line 310 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/33#-Kb8WhsQvaGdJ6Rkvrlt:-Kb9OyNIGFCk-IijcWsB:ba1yxel) ([raw file](https://github.com/sillsdev/lfmerge/blob/5210f6ff8b272a3301dca4bb88c008fd8f2e4a3c/src/LfMerge.Core.Tests/TestDoubles.cs#L310)):*

This is good, and should probably be cherry-picked even if we don't merge this PR.
</blockquote></details>

It's also only a workaround - looking at the writing systems after the data migration I see that some have a two-letter and some a three-letter abbreviation and data migrations seem to change them randomly. And when doing the data migration in FW we get yet different results. So I'm trying to find the root cause of this behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/33)*

ermshiperete:
This approach does not work with the current chorus code.

One problem is that it seems that this change leaves files in `data/testlangproj` in a state that hg sees as modified (line ending difference). While that would be easily fixable by amending the PR, a more serious problem is that LfMergeBridge when it splits the `.fwdata` file writes the XML files with `\r\n`. While hg ignores those line ending differences now it still creates an empty commit which causes the tests to fail.

We write `\r\n` because `Palaso.Xml.CanonicalXmlSettings.CreateXmlWriterSettings()` sets `NewLineChars` to `\r\n`. The reason for that is to make the linux implementation byte for byte compatible with the windows implementation.

This could be solved with the line ending normalization that this change introduces. Of course it would also need a change in Chorus or FlexBridge to always create a `.hgeol` file (and changing the value that `CreateXmlWriterSettings` sets). While that would work for all new repos we create, it will be a problem for existing repos: as soon as we create the `.hgeol` file all files appear as modified to hg, so that would need some additional thought and testing.

PR #37, opened 2017-03-02
Merge master into qa

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/37)


rmunn:
I think this PR was not what you intended: the comment says "Merge master into qa", but the PR is actually merging your copy of master into this repo's master (where there are four commits that already happened, though with different hashes -- presumably because they were cherry-picked). I'll close this in favor of #38, which is what I think you meant to do.

ermshiperete:
Selected wrong branches for this PR

PR #30, opened 2017-01-09
Remove queue file before running action

ermshiperete:
This fixes https://trello.com/c/HwvpS6w1.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/30)

PR #39, opened 2017-03-16
Dequeue projects before running LfMerge

rmunn:
This should ensure that LF doesn't get "stuck" thinking the project is pending. Now, what can happen is that LF's pending message disappears a second or so before the project starts the S/R process, but that's not a bad experience from the user's point of view.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/39)

### DIFF
--- a/src/LfMerge.Core/MainClass.cs
+++ b/src/LfMerge.Core/MainClass.cs
@@ -63,7 +63,7 @@ namespace LfMerge.Core
 			return containerBuilder;
 		}
 
-		public static void StartLfMerge(string projectCode, ActionNames action,
+		public static int StartLfMerge(string projectCode, ActionNames action,
 			string modelVersion, bool allowFreshClone, string configDir = null)
 		{
 			// Call the correct model version specific LfMerge executable
@@ -98,6 +98,7 @@ namespace LfMerge.Core
 				using (var process = Process.Start(startInfo))
 				{
 					process.WaitForExit();
+					return process.ExitCode;
 				}
 			}
 			catch (Exception e)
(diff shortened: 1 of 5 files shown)


megahirt:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMergeQueueManager/QueueManager.cs, line 56 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/39#-KfQE65rBP5hB64aH9xF:-KfQE65rBP5hB64aH9xG:b9lfwt8) ([raw file](https://github.com/sillsdev/lfmerge/blob/c243cba80a2459345b5927ae180f88ec7dde5252/src/LfMergeQueueManager/QueueManager.cs#L56)):*
> ```Smalltalk
> 							modelVersion, true);
> 
> 						// TODO: If LfMerge fails, should we re-queue the project, or not?
> ```

no, we should not requeue

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/39)*

ermshiperete:
I don't see much difference compared to #30 (except that we return an error code that we don't use), but that doesn't matter. :lgtm: 

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/39#-:-KfuAocnpms68IPV5F9Q:b-jc6cat)*

megahirt:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/39#-:-KfyWSh1f6yXXFpDBuFz:bnfp4nl)*

rmunn:
There's no real benefit to this one over #30, except that having the error code gives us some future flexibility (for example, if we ever decide that we *do* want to requeue projects under certain conditions). If I had realized ahead of time that this PR would essentially end up the same as #30, I would have just merged #30 instead.

PR #40, opened 2017-04-20
Fix Thai input system getting into non-Thai projects

rmunn:
The cause of the problem was a combination of two things:

1. The PHP code that creates a LexProjectModel creates two input systems, `en` and `th`, as "placeholders" so that its config can be correctly updated.

2. The "placeholder" input systems should normally be overwritten on initial clone. But when LfMerge detects an older project model and calls itself recursively, the "inner" LfMerge process sees an existing Mongo project and thinks that it contains user data that must be preserved, so it doesn't treat it as an initial clone. (This was a recent change intended to preserve user data if the `webwork/(project-code)` folder ever gets corrupted or deleted.) Then because it thinks there is valuable user data in Mongo, it does a TransferMongoToFdo first before doing the TransferFdoToMongo, and so the placeholder `th` input system is added to the FDO project. (Usually incorrectly: the bug report that triggered this fix was from a project not using Thai).

The solution I've implemented is to add a command-line parameter to LfMerge, which the "outer" LfMerge call can use to determine that this is actually an initial clone. If the `--initial-clone` parameter is passed, LfMerge will overwrite the existing Mongo config (including the placeholder input systems), and will skip the TransferMongoToFdo step, thus throwing away the data currently in Mongo because it's not actually real user data. (Obviously, using the `--initial-clone` parameter when there *is* valuable user data in Mongo would be a Bad Thing™, since that valuable user data could be lost.)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/40)

### DIFF
--- a/src/LfMerge.Core.Tests/Actions/EnsureCloneActionBridgeIntegrationTests.cs
+++ b/src/LfMerge.Core.Tests/Actions/EnsureCloneActionBridgeIntegrationTests.cs
@@ -26,8 +26,8 @@ namespace LfMerge.Core.Tests.Actions
 	{
 		private class EnsureCloneActionWithoutMongo: EnsureCloneAction
 		{
-			public EnsureCloneActionWithoutMongo(LfMergeSettings settings, ILogger logger, MongoProjectRecordFactory factory)
-				: base(settings, logger, factory)
+			public EnsureCloneActionWithoutMongo(LfMergeSettings settings, ILogger logger, MongoProjectRecordFactory factory, IMongoConnection connection)
+				: base(settings, logger, factory, connection)
 			{
 			}
 
@@ -58,7 +58,10 @@ namespace LfMerge.Core.Tests.Actions
 			Directory.CreateDirectory(LanguageDepotMock.ProjectFolderPath);
 			_lfProject = LanguageForgeProject.Create(TestLangProj);
 			_mongoProjectRecordFactory = MainClass.Container.Resolve<LfMerge.Core.MongoConnector.MongoProjectRecordFactory>() as MongoProjectRecordFactoryDouble;
-			_EnsureCloneAction = new EnsureCloneActionWithoutMongo(_env.Settings, _env.Logger, _mongoProjectRecordFactory);
+			// Even though the EnsureCloneActionWithoutMongo class has "WithoutMongo" in the name, the EnsureClone class which it inherits from
+			// needs an IMongoConnection argument in the constructor, so we have to create a MongoConnectionDouble that we're not going to use here.
+			IMongoConnection _mongoConnection = MainClass.Container.Resolve<IMongoConnection>();
+			_EnsureCloneAction = new EnsureCloneActionWithoutMongo(_env.Settings, _env.Logger, _mongoProjectRecordFactory, _mongoConnection);
 			LanguageDepotMock.Server = new MercurialServer(LanguageDepotMock.ProjectFolderPath);
 		}
 
--- a/src/LfMerge.Core.Tests/Actions/EnsureCloneActionTests.cs
+++ b/src/LfMerge.Core.Tests/Actions/EnsureCloneActionTests.cs
@@ -30,7 +30,7 @@ namespace LfMerge.Core.Tests.Actions
 			// Making a stub file so Chorus model.TargetLocationIsUnused will be false
 			File.Create(Path.Combine(_projectDir, "stub"));
 			_mongoProjectRecordFactory = MainClass.Container.Resolve<MongoProjectRecordFactory>() as MongoProjectRecordFactoryDouble;
-			_mongoConnection = MainClass.Container.Resolve<IMongoConnection>() as MongoConnectionDouble;
+			_mongoConnection = _mongoProjectRecordFactory.Connection as MongoConnectionDouble;
 			if (_mongoConnection == null)
 				throw new AssertionException("EnsureClone action tests need a mock MongoConnection that stores data in order to work.");
 
@@ -47,7 +47,7 @@ namespace LfMerge.Core.Tests.Actions
 		{
 			// for this test we don't want the test double for InternetCloneSettingsModel
 			_env.Dispose();
-			_env = new TestEnvironment(false);
+			_env = new TestEnvironment(registerSettingsModelDouble: false);
 
 			// Setup
 			var nonExistingProjectCode = Path.GetRandomFileName().ToLowerInvariant();
@@ -175,6 +175,80 @@ namespace LfMerge.Core.Tests.Actions
 			// Verify
 			Assert.That(Directory.Exists(Path.Combine(projectDir, ".hg")), Is.True,
 				"Didn't clone project");
+		}
+
+		[Test]
+		public void EnsureClone_ProjectThatHasNeverBeenCloned_RunsInitialClone()
+		{
+			// Setup
+			var projectDir = Path.Combine(_env.Settings.WebWorkDirectory, _projectCode);
+			Directory.CreateDirectory(projectDir);
+			var lfProject = LanguageForgeProject.Create(_projectCode);
+			var action = new EnsureCloneActionDoubleMockingInitialTransfer(_env.Settings, _env.Logger, _mongoProjectRecordFactory, _mongoConnection);
+			Assert.That(action.InitialCloneWasRun, Is.False);
+
+			// Execute
+			action.Run(lfProject);
+
+			// Verify
+			Assert.That(action.InitialCloneWasRun, Is.True);
+		}
+
+		[Test]
+		public void EnsureClone_ProjectThatHasAPreviouslyClonedDate_DoesNotRunInitialClone()
+		{
+			// Setup
+			var projectDir = Path.Combine(_env.Settings.WebWorkDirectory, _projectCode);
+			Directory.CreateDirectory(projectDir);
+			var lfProject = LanguageForgeProject.Create(_projectCode);
+			_mongoConnection.SetLastSyncedDate(lfProject, DateTime.UtcNow);
+			var action = new EnsureCloneActionDoubleMockingInitialTransfer(_env.Settings, _env.Logger, _mongoProjectRecordFactory, _mongoConnection);
+			Assert.That(action.InitialCloneWasRun, Is.False);
+
+			// Execute
+			action.Run(lfProject);
+
+			// Verify
+			Assert.That(action.InitialCloneWasRun, Is.False);
+		}
+
+		[Test]
+		public void EnsureClone_ProjectThatHasAPreviouslyClonedDateButDoesNotHaveAnSRProjectCode_RunsInitialClone()
+		{
+			// Setup
+			var projectDir = Path.Combine(_env.Settings.WebWorkDirectory, _projectCode);
+			Directory.CreateDirectory(projectDir);
+			var lfProject = LanguageForgeProject.Create(_projectCode);
+			_mongoConnection.SetLastSyncedDate(lfProject, DateTime.UtcNow);
+			var projectRecord = _mongoProjectRecordFactory.Create(lfProject);
+			projectRecord.SendReceiveProjectIdentifier = null;
+			var action = new EnsureCloneActionDoubleMockingInitialTransfer(_env.Settings, _env.Logger, _mongoProjectRecordFactory, _mongoConnection);
+			Assert.That(action.InitialCloneWasRun, Is.False);
+
+			// Execute
+			action.Run(lfProject);
+
+			// Verify
+			Assert.That(action.InitialCloneWasRun, Is.True);
+		}
+
+		[Test]
+		public void EnsureClone_ProjectThatHasPreviousUserData_DoesNotRunInitialClone()
+		{
+			// Setup
+			var projectDir = Path.Combine(_env.Settings.WebWorkDirectory, _projectCode);
+			Directory.CreateDirectory(projectDir);
+			var lfProject = LanguageForgeProject.Create(_projectCode);
+			var data = new SampleData();
+			_mongoConnection.UpdateMockLfLexEntry(data.bsonTestData);
+			var action = new EnsureCloneActionDoubleMockingInitialTransfer(_env.Settings, _env.Logger, _mongoProjectRecordFactory, _mongoConnection);
+			Assert.That(action.InitialCloneWasRun, Is.False);
+
+			// Execute
+			action.Run(lfProject);
+
+			// Verify
+			Assert.That(action.InitialCloneWasRun, Is.False);
 		}
 	}
 }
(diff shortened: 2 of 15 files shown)


rmunn:
I'm not 100% happy with the approach I took here, shoehorning the `TreatAsInitialClone` property into `ChorusHelper`. This feels like I'm making the class play two roles, and I'm introducing another global variable, increasing the potential complexity of the code and the difficulty of making future changes. (It's really a property on the `ChorusHelper` singleton, but singletons are essentially global variables in fancy dress, so it has all the same code-complexity drawbacks as a global variable).

However, I haven't yet found a better way of writing this, so I'm submitting this bugfix as PR for review. It does fix the user-facing bug; I've verified that by local testing on my machine. I don't have unit tests for this bugfix yet -- that's the next step -- but I might as well put it out for review now, while I start writing appropriate unit tests.

rmunn:
The Jenkins build fails because it's trying to build `web-languageforge` first as a dependency, and `web-languageforge` can't build because some of its own PHP dependencies require PHP 7.0, and the PHP version on the Jenkins build agents is PHP 5.6.22. This is not a failure that will be resolved quickly, I think, so it's probably better to review this PR without waiting for the Jenkins build to succeed.

ermshiperete:
The `lfmerge` builds on Jenkins build on Debian Jessie which has only PHP 5. We should update to Ubuntu Xenial which comes with PHP 7.

Re the changes here: I'm wondering: can't we check the SRState and if it's `SendReceiveStates.CLONING` it would tell us that it's an initial clone? Of course we would have to check that before we set it to CLONING again in `EnsureCloneAction.DoRun`.

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge.Core/Actions/Infrastructure/ChorusHelper.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-Ki9N_vkIPPzS5SdlmeY:-Ki9N_vl3jjto88zeJMA:b-wvb48) ([raw file](https://github.com/sillsdev/lfmerge/blob/c71e7489950a19f630d5c0ba3a4ad21734781a63/src/LfMerge.Core/Actions/Infrastructure/ChorusHelper.cs#L35)):*
> ```Smalltalk
> 		}
> 
> 		public static void SetTreatAsInitialClone(bool treatAsInitialClone)
> ```

I'm wondering why did you make this a method rather than a setter on `ThisIsAnInitialClone`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-Ki9PZaAlZQsvtqbxM54:bhv9gks)*

rmunn:
I looked at the SRState, and decided that it wasn't the right approach since there's plenty of other code around that feels free to play with it; in particular, we set it to IDLE at the end of each run and I wasn't sure I wanted to mess with that. But perhaps I was wrong about how much other code there is that sets SRState; I didn't write that part of the logic (I think you wrote that), and so that part of the design is less clear in my head. If using SRState to communicate "This is an initial clone" to the "inner" LfMerge would work, then that might be a good approach. But we'd want to make sure that we can tell this scenario (initial clone that's been "forwarded" to the inner LfMerge) apart from the "user data in Mongo that needs preserving" scenario. In the latter case, the `webwork` folder has somehow gotten lost, but there's still user data in Mongo from someone who was using LF in the past day/week/month/whatever. That user data in Mongo does not yet exist anywhere else, and will be lost forever if we overwrite that Mongo entry with one from FDO, so we do want to be careful to preserve it. So I tried to make sure that I could not *accidentally* enter the "initial clone forwarded to inner LfMerge" code path; it can only happen with a command-line option, rather than with a state file that other code can sometimes rewrite.

But as I said, the design you just suggested may indeed be the best one. If we can be sure we won't overwrite Mongo in a "`webwork` folder missing" scenario, then great: let's communicate that information via SRState, which *is* the most natural way to express it.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge.Core/Actions/Infrastructure/ChorusHelper.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-Ki9N_vkIPPzS5SdlmeY:-Ki9YmL54iZZ7VOIjalj:b8iro5w) ([raw file](https://github.com/sillsdev/lfmerge/blob/c71e7489950a19f630d5c0ba3a4ad21734781a63/src/LfMerge.Core/Actions/Infrastructure/ChorusHelper.cs#L35)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

I'm wondering why did you make this a method rather than a setter on `ThisIsAnInitialClone`?
</blockquote></details>

Because I was copying the design of SetModelVersion, where the property is an instance variable but there are static getter and setter methods that work with the singleton instance. I didn't put any design thought into this, I just copied the design of SetModelVersion, because the needs for both are pretty much the same. (The same code paths need to know about the model version *and* the initial-clone status).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-Ki9XM06nVsQXb5_BGH-:bcaeohr)*

ermshiperete:
Good thoughts. There are two instances where we set SRState to CLONING that are potentially dangerous: we set SRState to CLONING if the state file doesn't exist (in EncuresCloneAction.Pre...), and if we don't find a .fwdata file we delete the directory and clone again (in LfMerge/Program.cs).  So I don't think my suggestion will work.

Another thought: we want to overwrite the mongo data exactly once, and that is after the very first clone. Do we have any data in mongo that could tell us if we'd synced before, like last sr timestamp? When we do the initial clone this should be unset and so we could probably use that for detection.

(Otherwise this PR is fine, but I'm with you that I'm not 100% happy with this approach. Seems a little clumsy and it feels like there should be a way somewhere to detect that this is an initial clone rather than having to pass a flag around)

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-Ki9bz0ASZfyetfB7WEf:b9hepml)*

rmunn:
There *was* a way to detect if it's an initial clone -- the Mongo DB doesn't have an SR project code in it yet. (See the [`MongoProjectAlreadyExistsAndIsSRProject`](https://github.com/sillsdev/LfMerge/blob/c71e7489950a19f630d5c0ba3a4ad21734781a63/src/LfMerge.Core/Actions/EnsureCloneAction.cs#L215) function.) But that turned out to only be the case in the "outer" LfMerge instance, and not in the "inner" instance since at some point the SR project code gets set. Rather than hunt for the *right* way to do it and take more time away from the current common-good sprint, I opted for a quick-but-slightly-"dirty" solution, passing the flag around, instead.

After the sprint, I should probably revisit this and try to come up with a cleaner design.

megahirt:
What about checking to see if the LF project has any entries?  If number of
entries in mongo == 0, then overwrite mongo with initial clone.  If entries
> 0, then preserve mongo before transfer.

megahirt:
Also, this is high priority.  Let's do this right and then get back to the
common sprint.  A quick and dirty now approach is going to cost us more in
the long run.  Don't you agree?

rmunn:
Checking if the LF project has any entries would be a good way to proceed, but there are two issues with that approach:

1. It would require the `EnsureCloneAction` to take a dependency on `IMongoConnection`, which we have so far avoided having to have. (That's one reason why I didn't go with that approach when first checking if the project exists).
1. There *could* be user data embedded in the config settings in a real project whose `webwork` folder got lost. E.g., the user clones his empty FW project in LF, then sets up the view settings the way he wants them before letting his non-techie language helpers log in. When the user is done setting up those view settings, he clicks the Send/Receive button because he's made that a habit -- it's like clicking "File -> Save", and he trusts that nothing bad will happen if he does so. But oops, because there were no entries yet, we decided to overwrite his config. Now he doesn't trust the Send/Receive button any more.

I think the `LastSyncedDate` field in the project record might be a better one to check than the number of entries. But I'll go make **absolutely sure** that it can't end up being `null` in a "we have real user data" situation. Maybe I need to double it up with a check on the # of entries, and take that `IMongoConnection` dependency after all.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-KiDSs4kiWiu8WKH75l6:b-bhibsj)*

megahirt:
Taking the IMongoConnection dependency seems very reasonable given that
LFMerge is tightly coupled with mongodb.  I like the LastSyncedDate idea
best.  There may, in the future, be cases where we are doing a s/r sync
with an empty project (ie. no entries). I think that's a practically non
existent case right now.

rmunn:
If we take *just* the `LastSyncedDate` field as the condition to check, we won't need the MongoConnection dependency, as that field lives in the project record, which is taken care of by the MongoProjectRecordFactory dependency that already exists.

However, I like the belt-and-suspenders approach to making sure we don't lose user data, so I'll keep the `LastSyncedDate` check *and* the "more than zero lexical entries" check. (And OR them together, so that if either check is true, we won't overwrite Mongo data).

With those two checks in place, I feel safe removing the `--initial-clone` command-line parameter, which takes care of Eberhard's concerns (which I share) about that solution not feeling "right". I'll push that commit, and then I'll be happy with the state of this PR.

ermshiperete:
One minor suggestion, otherwise looks good.

---

Reviewed 8 of 9 files at r2, 1 of 6 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge.Core/MongoConnector/MongoConnection.cs, line 137 at r3](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiEcqm5NkDcsK0WhF1I:-KiEcqm6uKkTCWrvADPW:b-qi0tnk) ([raw file](https://github.com/sillsdev/lfmerge/blob/339d0ad9837412ed341a60dcc041e308c09ba11e/src/LfMerge.Core/MongoConnector/MongoConnection.cs#L137)):*
> ```Smalltalk
> 		}
> 
> 		public long EntryCount(ILfProject project)
> ```

A better name might be `LexiconEntryCount`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-KiEd1za6xQXIXR1SEff:bh5qq6v)*

rmunn:
If anyone wants to give this an LGTM stamp so GitHub lets me push the "Squash and Merge" button, I'll merge this into `master` now that it has working unit tests. I could also do that manually at the Git command line, but doing it through GitHub gives a cleaner history (e.g., it'll automatically sync up with the issue report).

---

Review status: 4 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMerge.Core/MongoConnector/MongoConnection.cs, line 137 at r3](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiEcqm5NkDcsK0WhF1I:-KiU3uhLosCxeTawKESm:bkwwk8g) ([raw file](https://github.com/sillsdev/lfmerge/blob/339d0ad9837412ed341a60dcc041e308c09ba11e/src/LfMerge.Core/MongoConnector/MongoConnection.cs#L137)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

A better name might be `LexiconEntryCount`
</blockquote></details>

I went with `LexEntryCount` since that matches the naming scheme used in the rest of the code (e.g., `LfLexEntry`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-KiU45ugb6n4IsIe94Eu:b-saguko)*

ermshiperete:
:lgtm:

---

Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/LfMerge.Core.Tests/TestDoubles.cs, line 312 at r4](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiV8sebT-G1JiTyKMtg:-KiV8secmnb4t2y95rrA:b5cksek) ([raw file](https://github.com/sillsdev/lfmerge/blob/ae9fbd6c48d28efa5cbb8c245caa50a657385d21/src/LfMerge.Core.Tests/TestDoubles.cs#L312)):*
> Memoize

Is this a typo or a word that I don't know?

---

*[src/LfMerge.Core.Tests/Actions/Infrastructure/ChorusHelperTests.cs, line 19 at r4](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiV9mOeNnmujYWfR-nt:-KiV9mOeNnmujYWfR-nu:b-fiuy3x) ([raw file](https://github.com/sillsdev/lfmerge/blob/ae9fbd6c48d28efa5cbb8c245caa50a657385d21/src/LfMerge.Core.Tests/Actions/Infrastructure/ChorusHelperTests.cs#L19)):*
> ```Smalltalk
> 		public void Setup()
> 		{
> 			_env = new TestEnvironment(registerSettingsModelDouble: false);
> ```

:+1:  I like this - makes it more clear what the 'false' does

---

*[src/LfMerge.Core/MongoConnector/MongoConnection.cs, line 137 at r3](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiEcqm5NkDcsK0WhF1I:-KiVA0Bp2ksqw77u3Aut:b-8fx2u7) ([raw file](https://github.com/sillsdev/lfmerge/blob/339d0ad9837412ed341a60dcc041e308c09ba11e/src/LfMerge.Core/MongoConnector/MongoConnection.cs#L137)):*

I went with `LexEntryCount` since that matches the naming scheme used in the rest of the code (e.g., `LfLexEntry`).
</blockquote></details>

:+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-:-KiVAF3OwAxHRsIxawzf:bnfp4nl)*

rmunn:

---

*[src/LfMerge.Core.Tests/TestDoubles.cs, line 312 at r4](https://reviewable.io:443/reviews/sillsdev/lfmerge/40#-KiV8sebT-G1JiTyKMtg:-KiVO6bsX0L4ICnD2uNW:b-opsaff) ([raw file](https://github.com/sillsdev/lfmerge/blob/ae9fbd6c48d28efa5cbb8c245caa50a657385d21/src/LfMerge.Core.Tests/TestDoubles.cs#L312)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

> Memoize

Is this a typo or a word that I don't know?
</blockquote></details>

It's [not a typo](https://en.wikipedia.org/wiki/Memoization). Although technically I'm not using memoization in the *strictest* sense, because the word memoization is usually reserved for caching the output of "pure" functions, where the output will always be the same value. Here, I'm explicitly relying on the fact that the output is mutable, and making sure to always return the same *object* when called with the same parameters, so that it can better simulate storing and retrieving values in the Mongo database. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/40)*

PR #41, opened 2017-06-21
Send/Receive Chorus Notes with LF comments

rmunn:
The corresponding [LfMergeBridge PR](https://github.com/sillsdev/flexbridge/pull/149) should be merged in first, then built, before this PR is merged. That will allow the TeamCity build process to obtain an LfMergeBridge DLL that knows the new GetChorusNotes and WriteToChorusNotes actions. If this PR is merged too soon, the resulting LfMerge.exe will run, but will throw an exception when you try to do a Send/Receive and it tries to handle comments.

There are no unit tests yet, but I'm creating the PR now so that we can see how it builds on TeamCity, and so that the code can be reviewed while we wait for the LfMergeBridge PR to be merged.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/41)


rmunn:
Closing and reopening to trigger another build.

ermshiperete:
Changes merged in 57496773c7dd26d8839e5bfe8772aa71ee4abcc4

PR #43, opened 2017-09-15
Only real changes should update DateModified

rmunn:
Specifically, if the fields in an LfOptionList haven't been modified at all, its DateModified shouldn't be updated either.

This has not been causing any bugs so far, but it does cause useless patches (where nothing has changed in Mongo except the DateModified value on option lists) to show up when using the lfmerge-autosrtests tool. With this change, that tool only creates patches if some actual data has changed in Mongo.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/43)

### DIFF
--- a/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs
+++ b/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs
@@ -47,7 +47,7 @@ namespace LfMerge.Core.DataConverters
 
 		public virtual LfOptionList PrepareOptionListUpdate(ICmPossibilityList fdoOptionList)
 		{
-			bool origLfOptionListModified = false;
+			bool optionListDiffersFromOriginal = false;
 			Dictionary<Guid, ICmPossibility> fdoOptionListByGuid = fdoOptionList.ReallyReallyAllPossibilities
 				// .Where(poss => poss.Guid != null) // Not needed as ICmPossibility GUIDs are not nullable
 				.ToDictionary(poss => poss.Guid);
@@ -58,15 +58,15 @@ namespace LfMerge.Core.DataConverters
 				if (_lfOptionListItemByGuid.TryGetValue(poss.Guid, out correspondingItem))
 				{
 					bool thisItemWasModified = SetOptionListItemFromCmPossibility(correspondingItem, poss);
-					origLfOptionListModified |= thisItemWasModified;
-					// Two-step process because we don't want || to short-circuit calling SetOptionListItemFromCmPossibility().
+					optionListDiffersFromOriginal |= thisItemWasModified;
+					// Two-step process because we don't want |= to short-circuit calling SetOptionListItemFromCmPossibility().
 					// QUESTION for code reviewers: would that intent have been clear if I had written the following line instead?
-					// origLfOptionListModified = SetOptionListItemFromCmPossibility(correspondingItem, poss) || origLfOptionListModified;
+					// optionListDiffersFromOriginal = SetOptionListItemFromCmPossibility(correspondingItem, poss) || optionListDiffersFromOriginal;
 				}
 				else
 				{
 					correspondingItem = CmPossibilityToOptionListItem(poss);
-					origLfOptionListModified = true;
+					optionListDiffersFromOriginal = true;
 				}
 				_lfOptionListItemByGuid[poss.Guid] = correspondingItem;
 				_lfOptionListItemByStrKey[correspondingItem.Key] = correspondingItem;
@@ -79,7 +79,13 @@ namespace LfMerge.Core.DataConverters
 				.Where(item => fdoOptionListByGuid.ContainsKey(item.Guid.GetValueOrDefault()))
 				.ToList();
 
-			if (origLfOptionListModified)
+			if (lfNewOptionList.Items.Count != _lfOptionList.Items.Count)
+			{
+				// Deleted at least one item because it had disappeared from FDO
+				optionListDiffersFromOriginal = true;
+			}
+
+			if (optionListDiffersFromOriginal)
 			{
 				lfNewOptionList.DateModified = DateTime.Now;  // TODO: Investigate why this was changed from UtcNow
 			}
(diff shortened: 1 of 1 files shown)


rmunn:
Jenkins build is failing with an unrelated error, BTW.

ermshiperete:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 64 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3ZtubwY5LuNbZ-k4n:-Ku3ZtucaXw20kFZcvEP:b-xlr94n) ([raw file](https://github.com/sillsdev/lfmerge/blob/10a9a76efe89517f1d8e1d4e279b0f82ebed17cc/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L64)):*
> ```Smalltalk
> 					// Two-step process because we don't want || to short-circuit calling SetOptionListItemFromCmPossibility().
> 					// QUESTION for code reviewers: would that intent have been clear if I had written the following line instead?
> 					// origLfOptionListModified = SetOptionListItemFromCmPossibility(correspondingItem, poss) || origLfOptionListModified;
> ```

The explicit comment makes it very clear why we have to do it that way. However, comments are often overlooked. The commented || line makes the intent a little less obvious, but it prevents accidental unwanted refactoring, so I'd prefer the commented || version.

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 156 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3cIXc-r1EsxCGgggI:-Ku3cIXdX6dHLMM_Vzge:b-ohnr6e) ([raw file](https://github.com/sillsdev/lfmerge/blob/10a9a76efe89517f1d8e1d4e279b0f82ebed17cc/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L156)):*
> ```Smalltalk
> 			var result = new LfOptionList();
> 			result.Items = new List<LfOptionListItem>();
> 			result.DateCreated = result.DateModified = DateTime.Now;  // TODO: Investigate why this was changed from UtcNow
> ```

Yes, that's very strange. It got introduced in [b35faad8](https://github.com/sillsdev/LfMerge/commit/b35faad8d9c5c4c242e15e07f381ab5cbecf9602) since FDO seemingly uses local time - which is so wrong.

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3doqt2zH9XsBfm8Fb:-Ku3doqt2zH9XsBfm8Fc:b-nttu98) ([raw file](https://github.com/sillsdev/lfmerge/blob/6023a168d6ac242a1886a9faf1ece6b9c86dc60e/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L82)):*
> _lfOptionList.Items

Will this always be not null?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-:-Ku3dzwICKvSBWbIZiLD:bnfp4nl)*

rmunn:
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 64 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3ZtubwY5LuNbZ-k4n:-KuIXwNdnT2yX2C9pRxk:b32w0x) ([raw file](https://github.com/sillsdev/lfmerge/blob/10a9a76efe89517f1d8e1d4e279b0f82ebed17cc/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L64)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

The explicit comment makes it very clear why we have to do it that way. However, comments are often overlooked. The commented || line makes the intent a little less obvious, but it prevents accidental unwanted refactoring, so I'd prefer the commented || version.
</blockquote></details>

OK

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 156 at r1](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3cIXc-r1EsxCGgggI:-KuIYLZaHg-iIqOgBnhh:b-8j5v9r) ([raw file](https://github.com/sillsdev/lfmerge/blob/10a9a76efe89517f1d8e1d4e279b0f82ebed17cc/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L156)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

Yes, that's very strange. It got introduced in [b35faad8](https://github.com/sillsdev/LfMerge/commit/b35faad8d9c5c4c242e15e07f381ab5cbecf9602) since FDO seemingly uses local time - which is so wrong.
</blockquote></details>

I'm going to leave it as local time to minimize changes in this PR, but I plan to switch it to UTC time later on since my tests so far show that Mongo is converting it to UTC anyway, and it's better for the code to explicitly document what's going to end up happening (UTC will be stored in Mongo). When I do switch this to UTC, I'll write an S/R test to verify that that change doesn't cause problems.

---

*[src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs, line 82 at r2](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-Ku3doqt2zH9XsBfm8Fb:-KuIYA8cTezVDof3jXcE:b-jeesat) ([raw file](https://github.com/sillsdev/lfmerge/blob/6023a168d6ac242a1886a9faf1ece6b9c86dc60e/src/LfMerge.Core/DataConverters/ConvertFdoToMongoOptionList.cs#L82)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

> _lfOptionList.Items

Will this always be not null?
</blockquote></details>

Yes; the `LfOptionList` constructor guarantees that `Items` will always be a non-null list, though it might be an empty list.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/43)*

ermshiperete:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/43#-:-KuI_1FoaT2_vX08HKtl:bnfp4nl)*

PR #45, opened 2018-02-20
Add ContextGuid property (LF-48)

ermshiperete:
This change also requires a change in flexbridge ([#202](https://github.com/sillsdev/flexbridge/pull/202))

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/45)


megahirt:
:lgtm: Everyone should be aware that Damien has already started creating .Net Core classes that essentially duplicates some of what we already have in LFMerge.  We should talk about how to bring these C# classes together so that we're not duplicating classes in both LFMerge and the C# API @ddaspit is developing

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/lfmerge/45#-:-L5oLy8qd49san-3i439:b-7ftzqj)*

PR #50, opened 2018-04-10
Update to Bugsnag v2 API

ermshiperete:
The v1 API will be shutdown on 2018-07-10, so we'll have to use the
new one which means to update the bugsnag nuget package we use.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/50)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build/NuGet.exe
+build/nuget.exe
 packages/
 bin/
 obj/
--- a/build/NuGet.targets
+++ b/build/NuGet.targets
@@ -9,7 +9,7 @@
 		<!-- Determines if package restore consent is required to restore packages -->
 		<RequireRestoreConsent Condition=" '$(RequireRestoreConsent)' != 'false' ">false</RequireRestoreConsent>
 
-		<!-- Download NuGet.exe if it does not already exist -->
+		<!-- Download nuget.exe if it does not already exist -->
 		<DownloadNuGetExe Condition=" '$(DownloadNuGetExe)' == '' ">true</DownloadNuGetExe>
 	</PropertyGroup>
 
@@ -24,14 +24,14 @@
 		<NuGetToolsPath>$(MSBuildThisFileDirectory)</NuGetToolsPath>
 
 		<!-- NuGet command -->
-		<NuGetExePath Condition=" '$(NuGetExePath)' == '' ">$(NuGetToolsPath)/NuGet.exe</NuGetExePath>
+		<NuGetExePath Condition=" '$(NuGetExePath)' == '' ">$(NuGetToolsPath)/nuget.exe</NuGetExePath>
 
 		<PackageSources Condition=" $(PackageSources) == '' ">@(PackageSource)</PackageSources>
 
 		<NuGetCommand Condition=" '$(OS)' == 'Windows_NT'">"$(NuGetExePath)"</NuGetCommand>
-		<NuGetCommand Condition=" '$(OS)' != 'Windows_NT' ">mono --runtime=v4.0.30319 $(NuGetExePath)</NuGetCommand>
+		<NuGetCommand Condition=" '$(OS)' != 'Windows_NT' ">$(NuGetToolsPath)/nuget</NuGetCommand>
 
-		<NuGetDownload>https://www.dropbox.com/s/j8kvxbg9506p21j/NuGet.exe</NuGetDownload>
+		<NuGetDownload>https://dist.nuget.org/win-x86-commandline/latest/nuget.exe</NuGetDownload>
 	</PropertyGroup>
 
 	<Target Name="CheckPrerequisites">
@@ -77,9 +77,9 @@
 				try {
 					OutputFilename = Path.GetFullPath(OutputFilename);
 
-					Log.LogMessage("Downloading latest version of NuGet.exe...");
+					Log.LogMessage("Downloading latest version of nuget.exe...");
 					WebClient webClient = new WebClient();
-					webClient.DownloadFile("https://nuget.org/NuGet.exe", OutputFilename);
+					webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/latest/nuget.exe", OutputFilename);
 
 					return true;
 				}
--- a/build/nuget
+++ b/build/nuget
@@ -1,0 +1,18 @@
+#!/bin/bash
+# This file is a workaround for using the current nuget.exe (which works with nuget packages that
+# specify dependencies for multiple frameworks, but requires a newer mono version) and having to
+# still use mono 3.x for building and running the project.
+
+if [ -f /opt/mono5-sil/bin/mono ]; then
+	export MONO_PREFIX=/opt/mono5-sil
+elif [ -f /opt/mono4-sil/bin/mono ]; then
+	export MONO_PREFIX=/opt/mono4-sil
+else
+	echo "Need mono4-sil or mono5-sil installed to be able to restore the packages"
+	exit 1
+fi
+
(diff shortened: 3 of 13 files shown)


rmunn:
:lgtm:

I admit that I skimmed through the ExceptionLogging.cs file rather than taking the time to understand it in detail, but it looks right, and I think I'd like to see us merge this in before BugSnag takes their v1 API endpoint down. If there are any problems in here, they should turn up when we test lfmerge on the dev and QA servers before deploying to live.

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved.

---

*[build/nuget, line 4 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXTdLg0YmWkaOnYFGp:-LAXTdLh4jVSKDtVb8Gr:b-5f634i) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/build/nuget#L4)):*
> ```
> # This file is a workaround for using the current nuget.exe (which works with nuget packages that
> # specify dependencies for multiple frameworks, but requires a newer mono version) and having to
> # still use mono 3.x for building and running the project.
> ```

BTW, is there any way for us to use Mono 4 or 5 to build LfMerge? Or is the requirement for Mono 3 due to some of the details of how FieldWorks is built? If it would be possible to get LfMerge building with mono4-sil or mono5-sil, that would allow us to use C# 6 or C# 7 features, which in some cases would make the logic a little cleaner. But don't spend too much time on that idea if it's difficult or time-consuming.

---

*[build/NuGet.targets, line 34 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXUHsgDXRFtpnmkaij:-LAXUHsgDXRFtpnmkaik:b68n1vi) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/build/NuGet.targets#L34)):*
> ```XML
> 		<NuGetCommand Condition=" '$(OS)' != 'Windows_NT' ">$(NuGetToolsPath)/nuget</NuGetCommand>
> 
> 		<NuGetDownload>https://dist.nuget.org/win-x86-commandline/latest/nuget.exe</NuGetDownload>
> ```

Happy to see the all-lowercase nuget.exe here; the first time I started working on C# on Linux, I had a lot of trouble with tools that had just assumed a case-insensitive filesystem, and had inconsistent spellings of NuGet.exe (I remember seeing it spelled both NuGet.exe and nuget.exe in the *same file* once, which obviously would only work on Windows, though I don't remember now what file that was).

---

*[src/LfMerge.Core/Logging/ExceptionLogging.cs, line 49 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXV9qG4lMzv1zFlI20:-LAXV9qG4lMzv1zFlI21:bij6rhq) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/src/LfMerge.Core/Logging/ExceptionLogging.cs#L49)):*
> ```Smalltalk
> 		}
> 
> 		private static string GetOSInfo()
> ```

Is this logic something that could be pulled out into a class in LibPalaso? Or does LibPalaso's OS detection have other needs, that aren't similar enough to this function to be worth extracting?

Actually, now that I think about it, a class for integrating with BugSnag for exception handling might be a nice thing to put into LibPalaso eventually, maybe as an optional thing (a separate DLL? A separate NuGet package?) so apps using LibPalaso that don't want BugSnag integration don't have to use it. Food for thought in the future...

---

*[src/LfMerge.Core/Logging/ExceptionLogging.cs, line 162 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXVu-b35m179d4ictJ:-LAXVu-b35m179d4ictK:b-bifpzm) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/src/LfMerge.Core/Logging/ExceptionLogging.cs#L162)):*
> ```Smalltalk
> 			try
> 			{
> 				using (var searcher = new ManagementObjectSearcher("SELECT * FROM Win32_OperatingSystem"))
> ```

I'm not familiar with ManagementObjectSearcher at all. Learned something new today. I assume this would fail completely on Linux, but of course this function only gets called if the OS has reported itself as Win32NT.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-:-LAXWNcK-SW3jCZOtUZw:boiv4nw)*

ermshiperete:
Review status: 0 of 13 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[build/nuget, line 4 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXTdLg0YmWkaOnYFGp:-LAXY9Jc6SZfk2vCvdk4:b-ct9y9p) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/build/nuget#L4)):*

BTW, is there any way for us to use Mono 4 or 5 to build LfMerge? Or is the requirement for Mono 3 due to some of the details of how FieldWorks is built? If it would be possible to get LfMerge building with mono4-sil or mono5-sil, that would allow us to use C# 6 or C# 7 features, which in some cases would make the logic a little cleaner. But don't spend too much time on that idea if it's difficult or time-consuming.
</blockquote></details>

I wish we could do that. Unfortunately there's not much we can do without quite a bit of work. FW 8.3 uses mono 3, and getting that to work with 4 or even 5 means several days of work, potentially also in libpalaso, chorus etc (for the branches lfmerge uses).

When we switch to support FW 9 we'll get mono 4 (or maybe even 5 if we invest a little more time).

---

*[src/LfMerge.Core/Logging/ExceptionLogging.cs, line 49 at r1](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-LAXV9qG4lMzv1zFlI20:-LAXZYDyB9mSPxrZUPFh:bafzbjf) ([raw file](https://github.com/sillsdev/lfmerge/blob/0c6a1401456a3f96c7c99c08e2fd3bca632187c0/src/LfMerge.Core/Logging/ExceptionLogging.cs#L49)):*

Is this logic something that could be pulled out into a class in LibPalaso? Or does LibPalaso's OS detection have other needs, that aren't similar enough to this function to be worth extracting?

Actually, now that I think about it, a class for integrating with BugSnag for exception handling might be a nice thing to put into LibPalaso eventually, maybe as an optional thing (a separate DLL? A separate NuGet package?) so apps using LibPalaso that don't want BugSnag integration don't have to use it. Food for thought in the future...
</blockquote></details>

Yes, I already put that in the master branch of libpalaso. But since we're using libpalaso-2.6 that's not available here...

Because of the dependency I didn't add this logic to libpalaso, but putting it in a separate DLL/package is a good idea for the future...

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50)*

rmunn:
:lgtm: in revision 2 as well.

---

Reviewed 11 of 13 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/lfmerge/50#-:-LAkD0hB2qW_fNZbVO12:bzhszd)*

rmunn:
Build appears to be timing out in the `DownloadDependencies` target, probably because the Jenkins build agent can't access TeamCity to download the dependencies. My guess is that a firewall configuration somewhere needs to be tweaked.

PR #67, opened 2018-09-26
Improve Chorus labels on comments for S/R (LF-198)

ermshiperete:
When we have a field-level comment we now display the fieldname,
writing system abbreviation and field value (if available) in the
Chorus field label to provide more context.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/67)


ermshiperete:
@megahirt Can you please review this if you agree with the implemented behaviour? See https://jira.sil.org/browse/LF-198 for a screenshot how things look with this change.

PR #74, opened 2018-12-03
 Adjust LfMerge to FieldWorks 9 

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/74)


megahirt:
Do the changes made to update the test data to 7000072 also leave some data at an earlier version so that the older version of FW can still be tested?

ermshiperete:
The `fieldworks8-*` branches have/will have the test data for earlier versions (FW8). `master, qa, and live` will only have the data for the latest model version.

Too many things have changed so that it would be difficult to continue to keep all versions in one branch.

megahirt:
Ok, this makes sense.  Will you package and deploy from both the FW 8
branch and live branch so that we can support both version 8 and 9
simultaneously?

ermshiperete:
Yes, that's the plan.

PR #88, opened 2019-07-16
Disable global writing system store (LF-297)

ermshiperete:
A global writing system store doesn't make sense in the context of
LfMerge where we deal with different independent projects. Disabling
it prevents that multiple projects share the same global store.
Part of a fix for LF-297.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/88)


ermshiperete:
Wrong branch - should do it on the `master` branch first.

PR #94, opened 2019-10-04
Do not convert LIFTid field in entries or senses (FW8 branch)

rmunn:
Language Forge doesn't need it if a GUID is present, and getting the LIFTid from FDO can cause issues if the project had import residue.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/94)


rmunn:

PR #98, opened 2020-06-04
Fix custom fields bug

rmunn:
When custom fields referenced a ListReference and there was no GUID in the Mongo data, and the custom field had a wsSelector that was a negative number (a "magic" ws), Send/Receive would crash. This fixes that bug and includes a regression test to ensure the bug stays fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/98)

### DIFF
--- a/src/LfMerge.Core.Tests/Actions/SynchronizeActionTests.cs
+++ b/src/LfMerge.Core.Tests/Actions/SynchronizeActionTests.cs
@@ -230,7 +230,7 @@ namespace LfMerge.Core.Tests.Actions
 		[Test]
 		public void SynchronizeAction_LFDataChangedLDDataChanged_LFWins()
 		{
-			//Setup
+			// Setup
 			TestEnvironment.CopyFwProjectTo(modifiedTestProjectCode, _lDSettings.WebWorkDirectory);
 			Directory.Move(Path.Combine(_lDSettings.WebWorkDirectory, modifiedTestProjectCode), LanguageDepotMock.ProjectFolderPath);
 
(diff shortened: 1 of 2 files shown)


src/LfMerge.Core.Tests/Actions/SynchronizeActionTests.cs
```diff
@@ -506,6 +506,83 @@ public void SynchronizeAction_LFDataChangedLDOtherDataChanged_ModifiedDateUpdate
...
+			_mongoConnection.UpdateRecord(_lfProject, lfEntry);
+
+			_lDProject = new LanguageDepotMock(testProjectCode, _lDSettings);
+			var lDcache = _lDProject.FieldWorksProject.Cache;
+			var lDLcmEntry = lDcache.ServiceLocator.GetObject(_testEntryGuid) as ILexEntry;
+			var data = (SIL.LCModel.Application.ISilDataAccessManaged)lDcache.DomainDataByFlid;
+			int ownedHvo = data.get_ObjectProp(lDLcmEntry.Hvo, listRef_flid);
+			if (ownedHvo == 0 || !data.get_IsValidObject(ownedHvo))
+			{
+				throw new Exception("Custom field value in test data was invalid during setup");
```
ermshiperete, line 552: It might be better to use `Assert.Fail` here instead of throwing an exception, or replace the `if` with something like:

```
Assert.That(ownedHvo, Is.Not.EqualTo(0), "Custom field value in test data was invalid during setup");
Assert.That(data.get_IsValidObject(ownedHvo), IsTrue, "Custom field value in test data was invalid during setup");
```

(similar below)

PR #100, opened 2020-06-30
Tweak unit tests for new LfMergeBridge branching

rmunn:
Now that LfMergeBridge uses ###.### style branch names, the unit tests need to be updated to match those expectations. This commit updates the testlangproj-modified project to include a 7500002.7000072 branch, so that the SynchronizeActionTests suite will pass with the new version of LfMergeBridge.

Unit tests will fail with current LfMergeBridge, but should pass once https://github.com/sillsdev/flexbridge/pull/285 is applied and LfMergeBridge is recompiled.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/100)


ermshiperete:
Do we still have to support model version `7000072`? Are there any released FW versions with that model version? If so we'd have to create a new branch with the previous code and create a package etc. - and make sure that things work. I don't remember, but will ask on slack.

Otherwise this change lgtm.

rmunn:
FLEx 8.3 uses 7000070. FLEX 9.0.7 beta used 7000072, but the 9.0.8 release went to the ###.### numbering scheme, so there are no released versions using 7000072 that I know of. If someone's using FLEx 9.0.7 beta, they probably won't mind upgrading to the final released version, so that's not an issue.

PR #106, opened 2021-04-27
Don't call string.Format twice

rmunn:
The logger code already calls string.Format on the message and parameters it receives. This is sometimes causing an error about "Input string was not in a correct format" in the second call, when the first string.Format call happens to produce output that contains formatting characters that the second string.Format thinks it should handle. The result is swallowed exceptions. By removing one call to string.Format, we should be able to see the actual exceptions.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/106)

### DIFF
--- a/src/LfMerge/Program.cs
+++ b/src/LfMerge/Program.cs
@@ -146,15 +146,15 @@ namespace LfMerge
 
 				if (project.State.SRState != ProcessingState.SendReceiveStates.ERROR)
 				{
-					MainClass.Logger.Error(string.Format(
+					MainClass.Logger.Error(
 						"Putting project '{0}' on hold due to unhandled exception: \n{1}",
-						projectCode, e));
+						projectCode, e);
 					if (project != null)
 					{
 						project.State.SetErrorState(ProcessingState.SendReceiveStates.HOLD,
-							ProcessingState.ErrorCodes.UnhandledException, string.Format(
+							ProcessingState.ErrorCodes.UnhandledException,
 								"Putting project '{0}' on hold due to unhandled exception: \n{1}",
-								projectCode, e));
+								projectCode, e);
 					}
 				}
 			}


megahirt (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/106)* status: 0 of 1 LGTMs obtained (waiting on @rmunn)

___
*[src/LfMerge/Program.cs, line 151 at r1](https://reviewable.io/reviews/sillsdev/lfmerge/106#-MZGaIEB-aA9gW5sHtHn:-MZGaIEB-aA9gW5sHtHo:bp7gcwu) ([raw file](https://github.com/sillsdev/lfmerge/blob/fa591fcc4b02074079b13ac81caae16211073cfd/src/LfMerge/Program.cs#L151)):*
> ```Smalltalk
> 					MainClass.Logger.Error(string.Format(
> 						"Putting project '{0}' on hold due to unhandled exception: \n{1}",
> 						projectCode, e));
> ```

You mentioned in your PR that you would try calling e.ToString() here, which seems to me the right thing to do.  But you've removed the string.Format() all together, which doesn't quite make sense to me given that you are passing in a formatted string.  But I don't actually see how this works.  If you say it works, well, it works.  Maybe you can explain it to me.

rmunn (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/106)* status: 0 of 1 LGTMs obtained (waiting on @megahirt)

___
*[src/LfMerge/Program.cs, line 151 at r1](https://reviewable.io/reviews/sillsdev/lfmerge/106#-MZGaIEB-aA9gW5sHtHn:-MZGbR8kCmtXO3RmV_QY:b-6keg6) ([raw file](https://github.com/sillsdev/lfmerge/blob/fa591fcc4b02074079b13ac81caae16211073cfd/src/LfMerge/Program.cs#L151)):*
<details><summary><i>Previously, megahirt (Christopher Hirt) wrote…</i></summary><blockquote>

You mentioned in your PR that you would try calling e.ToString() here, which seems to me the right thing to do.  But you've removed the string.Format() all together, which doesn't quite make sense to me given that you are passing in a formatted string.  But I don't actually see how this works.  If you say it works, well, it works.  Maybe you can explain it to me.
</blockquote></details>

The Logger.Error method calls string.Format, so calling it twice is not needed. However, this is not the actual problem (I read the stack trace wrong), so I'm going to close this PR after all.

rmunn:
The exception that led to me wanting this was:

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.FormatException: Input string was not in a correct format.
  at System.Text.StringBuilder.AppendFormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x000b2] in /build/mono5-sil-5.16.0.179/external/corefx/src/Common/src/CoreLib/System/Text/StringBuilder.cs:1446 
  at System.String.FormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x0000e] in /build/mono5-sil-5.16.0.179/external/corefx/src/Common/src/CoreLib/System/String.Manipulation.cs:471 
  at System.String.Format (System.String format, System.Object[] args) [0x00018] in /build/mono5-sil-5.16.0.179/external/corefx/src/Common/src/CoreLib/System/String.Manipulation.cs:436 
  at LfMerge.Core.Logging.LoggerBase.Error (System.String messageFormat, System.Object[] messageParts) [0x00000] in /build/lfmerge-7000072-8UUe80/lfmerge-7000072-2.0.55.21/src/LfMerge.Core/Logging/LoggerBase.cs:43 
  at LfMerge.Program.RunAction (System.String projectCode, LfMerge.Core.Actions.ActionNames currentAction) [0x00181] in /build/lfmerge-7000072-8UUe80/lfmerge-7000072-2.0.55.21/src/LfMerge/Program.cs:149 
  at LfMerge.Program.Main (System.String[] args) [0x00120] in /build/lfmerge-7000072-8UUe80/lfmerge-7000072-2.0.55.21/src/LfMerge/Program.cs:68 
```

Looking at the line numbers, I see that I had misread the final line as being line 168, but it's really line 68. Line 68 is the part where it gets the Chorus branch name and Chorus tries to parse it as an int, so it's the same old Chorus branch change popping up again. Closing this PR as it's not actually the right fix and not actually needed AFIACT.

PR #108, opened 2021-05-05
Fix exception in FLEx 8 project upgraded to FLEx 9

### LINKED ISSUE #107: Send/Receive on hold when a FLEx 8 project has FLEx 9 commits in Language Depot history
At some point in the FLEx 9 development process, the Chorus branch name format went from `7000072` to `7500002.7000072`. The version of Chorus used in lfmerge-7000072 is able to parse that branch format, but the versions of Chorus used in lfmerge-7000070 and earlier use `System.Int32.Parse` to parse branch names, which throws "System.FormatException: Input string was not in the correct format" and puts the project on hold.

We have code in LfMerge to detect that a newer model was pulled and re-run LfMerge with a later model version, but that code is being bypassed by this exception. There are two ways we could fix this issue:

1. Update older Chorus versions used by LfMerge 70000{68,69,70} so that they parse those branch names correctly. Or:

2. LfMerge catches System.FormatException and if it's from System.Int32.Parse, then assume that the model is `7500002.7000072` and just run lfmerge-7000072 to handle that Send/Receive.

The first is the "correct" fix, but the second is the quick fix that we can put into place much more easily. I lean towards doing the second now, and the first if/when we can find the time.

rmunn:
Due to an internal change in the Send/Receive system, when a FLEx 8 project upgrades to FLEx 9, LfMerge throws an unhandled exception in the Send/Receive process. This should fix that issue by forcing LfMerge to retry the Send/Receive with the FLEx 9 code if that happens.

Fixes #107.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/108)

### DIFF
--- a/src/LfMerge.Core.Tests/Actions/SynchronizeActionBridgeIntegrationTests.cs
+++ b/src/LfMerge.Core.Tests/Actions/SynchronizeActionBridgeIntegrationTests.cs
@@ -233,6 +233,73 @@ namespace LfMerge.Core.Tests.Actions
 			Assert.That(_lfProject.State.SRState, Is.EqualTo(ProcessingState.SendReceiveStates.SYNCING));
 		}
 
+		public void Success_NewBranchFormat_LfMerge68()
+		{
+			// Setup
+			var ldDirectory = CopyModifiedProjectAsTestLangProj(_lDSettings.WebWorkDirectory);
+			TestEnvironment.CopyFwProjectTo(TestLangProj, _env.Settings.WebWorkDirectory);
+			TestEnvironment.WriteTextFile(Path.Combine(_env.Settings.WebWorkDirectory, TestLangProj, "FLExProject.ModelVersion"), "{\"modelversion\": 7000068}");
+			LanguageDepotMock.Server.Start();
+			var oldHashOfLd = MercurialTestHelper.GetRevisionOfTip(ldDirectory);
+
+			// Execute
+			_synchronizeAction.Run(_lfProject);
+
+			// Verify
+			Assert.That(MercurialTestHelper.GetRevisionOfWorkingSet(_lfProject.ProjectDir),
+				Is.EqualTo(MercurialTestHelper.GetRevisionOfTip(ldDirectory)),
+				"Our repo doesn't have the changes from LanguageDepot");
+			Assert.That(MercurialTestHelper.GetRevisionOfTip(ldDirectory), Is.EqualTo(oldHashOfLd));
+			Assert.That(_env.Logger.GetErrors(), Is.Null.Or.Empty);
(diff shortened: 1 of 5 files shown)


megahirt (approved):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/108#-:-MZvmjTu25QtPvxHdIC8:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained (waiting on @rmunn)

___
*[src/LfMerge.Core/MagicStrings.cs, line 86 at r1](https://reviewable.io/reviews/sillsdev/lfmerge/108#-MZvloyhCgTTONYwl0H8:-MZvloyhCgTTONYwl0H9:bmty73v) ([raw file](https://github.com/sillsdev/lfmerge/blob/43f036c170d4056c6806db86ba2c3fa98528d5ff/src/LfMerge.Core/MagicStrings.cs#L86)):*
> ```Smalltalk
> 		// Field names to use in an LfCommentRegarding instance
> 		public const string LfFieldNameForDefinition = "definition";
> 		public const string LfFieldNameForExampleSentence = "definition";
> ```

So I was just browsing through the code and noticed this line which looks like a bug.  Should it not say "sentence" ?

`LfFieldNameForExampleSentence = "sentence";`

rmunn:
> So I was just browsing through the code and noticed this line which looks like a bug. Should it not say "sentence" ?
> 
> `LfFieldNameForExampleSentence = "sentence";`

Yes, that looks wrong. I'd have to check the Language Forge PHP code to see whether the field name is `sentence` or `example` in LF. I'll create a new issue to track this so that I don't forget about it.

rmunn:
Created https://github.com/sillsdev/LfMerge/issues/109 to track the probably-wrong string. I'm currently working on getting LfMerge PRs to build again on Jenkins (last build was 11 months ago), which will probably take me until tomorrow. Once I get a build and can test it, I'll merge this PR into master so it can be deployed.

rmunn:
Discovered that my bugfix so far isn't quite sufficient (see https://github.com/sillsdev/LfMerge/issues/107#issuecomment-850205243 for details). The new commits I just pushed should take care of that, where we'll now correctly extract DB version number 7000072 from a branch named `7500002.7000072`.

rmunn:
Note that #114 should be merged at the same time as this one; this one is against `master` and #114 is against `fieldworks-master`. Both branches need this bugfix.

rmunn:
Tested and working locally, and all Send/Receive unit tests passing. Merging. Once Jenkins builds the resulting packages, we can push them to QA and run them there for further testing.

PR #110, opened 2021-05-05
Bugfix: Correct example sentence magic string

### LINKED ISSUE #109: Wrong(?) field name in MagicStrings.cs
One magic string in MagicStrings.cs is `LfFieldNameForExampleSentence = "definition"`, and that should probably be `"sentence"` or `"example"`. I'll have to check the Language Forge code to be sure which one it is.

megahirt:
When reviewing code on a different PR, I noticed that the value of `LFFieldNameForExampleSentence` was set to `definition` when it most definitely should be set to `sentence`.  This looks wrong and is likely the result of a copy/paste error (from the line above).  I have no way to test this, and this PR doesn't fix a known bug.  However, it seems so obviously wrong that I got up the guts to create this PR :)

Fixes #109 

References in web-languageforge to support this PR:

- https://github.com/sillsdev/web-languageforge/blob/master/src/angular-app/languageforge/lexicon/shared/model/lex-example.model.ts#L7https://github.com/sillsdev/web-languageforge/blob/master/src/Api/Model/Languageforge/Lexicon/LexExample.php#L144
-

### DIFF
--- a/src/LfMerge.Core/MagicStrings.cs
+++ b/src/LfMerge.Core/MagicStrings.cs
@@ -83,7 +83,7 @@ namespace LfMerge.Core
 
 		// Field names to use in an LfCommentRegarding instance
 		public const string LfFieldNameForDefinition = "definition";
-		public const string LfFieldNameForExampleSentence = "definition";
+		public const string LfFieldNameForExampleSentence = "sentence";
 
 		// Fake language codes used in storing custom GenDate and int fields in Mongo
 		public const string LanguageCodeForGenDateFields = "qaa-Qaad";


rmunn (approved):
LGTM.

rmunn:
The one thing holding me back from merging this is that I want to check the Mongo database on QA (and live) to see if any migration script is needed. If we don't run a migration script, the worst that would happen is that some people's comments might show up in the wrong part of the entry, so that's not too bad.

rmunn:
I may be able to get to this one before I leave, but I'm marking it as "On hold" in case I don't.

PR #111, opened 2021-05-07
Update GitVersion tool and config

rmunn:
At some point recently, GitVersion changed the name of the `master` branch to `main`, and if we run `dotnet gitversion` with our current config we get a KeyNotFoundException with the message "The given key 'master' was not present in the dictionary." Changing `master` to `main` everywhere except the branch-name-matching regex fixes this.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/111)

PR #113, opened 2021-05-13
Use dotnet-gitversion tool in FW8 build

rmunn:
This is #111 but applied to the fieldworks8-master branch so that Jenkins can build that branch again.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/113)

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "gitversion.tool": {
+      "version": "5.6.9",
+      "commands": [
+        "dotnet-gitversion"
+      ]
+    }
+  }
+}
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,23 +1,31 @@
 mode: Mainline
+continuous-delivery-fallback-tag: dev
 branches:
   live:
     tag:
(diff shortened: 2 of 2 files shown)


rmunn:
The Jenkins build has succeeded at the GitVersion step, so this is safe to merge in even if the Jenkins build subsequently fails. I'm going to do so now.

rmunn:
By using Mono 5 for the build, and temporarily pinning a TeamCity build that produced artifacts, I've been able to get the fieldworks8-master build running again on Jenkins. There are a lot of unit test failures, but we can deal with those later. Mono 5 is required for the `moz-roots` step to work (using `mono-sil` tries to download certificate data from `http://mxr.mozilla.org/`, a server that no longer exists). And downloading FLExBridge DLLs from a pinned TeamCity build is required because the TeamCity FLExBridge builds of the `lfmerge` branch are currently building the DLLs successfully but then failing at a later step with a "stack smashing detected: /opt/mono5-sil/bin/mono terminated" message. I've been unable to resolve that issue as it appears to be a bug in Mono, but since it successfully built DLLs and we can use them for build dependencies, I pinned one of those builds despite TeamCity considering it a failed build.

So I'm going to merge this PR, as the point of it is to get the Jenkins build running again, not necessarily to get every unit test passing. The failing unit tests will be dealt with in a future PR.

PR #114, opened 2021-05-14
Fix exception in FLEx 8 project upgraded to FLEx 9

### LINKED ISSUE #107: Send/Receive on hold when a FLEx 8 project has FLEx 9 commits in Language Depot history

rmunn:
Due to an internal change in the Send/Receive system, when a FLEx 8 project upgrades to FLEx 9, LfMerge throws an unhandled exception in the Send/Receive process. This should fix that issue by forcing LfMerge to retry the Send/Receive with the FLEx 9 code if that happens.

Fixes #107.

**NOTE:** This is #108 applied to the correct branch (this fix needs to be in the older versions of LfMerge, and it doesn't hurt to have it on `master` as well so I'll leave #108 open).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/114)

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="TeamCity" value="http://build.palaso.org/guestAuth/app/nuget/v1/FeedService.svc/" />
+  </packageSources>
+</configuration>
--- a/build/LfMerge.dep
+++ b/build/LfMerge.dep
@@ -73,8 +73,8 @@ L10NSharp.pdb=>lib/
 
 [TC::FLExBridgeLfmergeLinux64Continuous]
 Name=FLEx Bridge-lfmerge-Linux64-Continuous
-RevisionName=lastSuccessful
-RevisionValue=latest.lastSuccessful
+RevisionName=lastPinned
+RevisionValue=latest.lastPinned
 Condition=Linux
 Path=LibFLExBridge-ChorusPlugin.*=>lib/
 LibTriboroughBridge-ChorusPlugin.*=>lib/
--- a/build/NuGet.targets
+++ b/build/NuGet.targets
@@ -13,12 +13,6 @@
 		<DownloadNuGetExe Condition=" '$(DownloadNuGetExe)' == '' ">true</DownloadNuGetExe>
 	</PropertyGroup>
 
-	<ItemGroup Condition=" '$(PackageSources)' == '' ">
-		<!-- Package sources used to restore packages. -->
-		<!-- The official NuGet package source (https://nuget.org/api/v2/) will be excluded if package sources are specified and it does not appear in the list -->
-		<PackageSource Include="https://nuget.org/api/v2/" />
-		<PackageSource Include="http://build.palaso.org/guestAuth/app/nuget/v1/FeedService.svc/" />
-	</ItemGroup>
 
 	<PropertyGroup>
 		<NuGetToolsPath>$(MSBuildThisFileDirectory)</NuGetToolsPath>
@@ -26,7 +20,7 @@
 		<!-- NuGet command -->
 		<NuGetExePath Condition=" '$(NuGetExePath)' == '' ">$(NuGetToolsPath)/nuget.exe</NuGetExePath>
 
-		<PackageSources Condition=" $(PackageSources) == '' ">@(PackageSource)</PackageSources>
+		<!-- <PackageSources Condition=" $(PackageSources) == '' ">@(PackageSource)</PackageSources> -->
 
 		<NuGetCommand Condition=" '$(OS)' == 'Windows_NT'">"$(NuGetExePath)"</NuGetCommand>
 		<NuGetCommand Condition=" '$(OS)' != 'Windows_NT' ">$(NuGetToolsPath)/nuget</NuGetCommand>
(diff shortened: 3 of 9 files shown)


jasonleenaylor (commented):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/114)* status: 0 of 1 LGTMs obtained (waiting on @rmunn)

___
*[src/LfMerge.Core/Actions/SynchronizeAction.cs, line 98 at r1](https://reviewable.io/reviews/sillsdev/lfmerge/114#-M_fTAcP0RL2j-PGuDSl:-M_fTAcP0RL2j-PGuDSm:b-6gpivb) ([raw file](https://github.com/sillsdev/lfmerge/blob/2cb994c9eba407ab9e1bfefcefad7844784083cc/src/LfMerge.Core/Actions/SynchronizeAction.cs#L98)):*
<details><summary><i alt="summary">Quoted 15 lines of code…</i></summary>

> ```
> 				try {
> 					if (!LfMergeBridge.LfMergeBridge.Execute("Language_Forge_Send_Receive", Progress,
> 						options, out syncResult))
> 					{
> 						Logger.Error(syncResult);
> 						return;
> 					}
> 				} catch (System.FormatException e) {
> 					if (e.StackTrace.Contains("System.Int32.Parse")) {
> 						ChorusHelper.SetModelVersion(MagicStrings.MinimalModelVersionForNewBranchFormat.ToString());
> 						return;
> 					} else {
> 						throw;
> 					}
> 				}
> ```
</details>

Braces are inconsistent, don't match the rest of the file, and also don't match general C# style guidelines. 

I'm not a developer on this tool so I can't speak with authority on this, but I would expect that you could fix the old version by just updating its version of chorus and avoid this exception entirely.

rmunn:
@jasonleenaylor wrote:
> I'm not a developer on this tool so I can't speak with authority on this, but I would expect that you could fix the old version by just updating its version of chorus and avoid this exception entirely.

I wish it was that simple, but LfMerge needs to use FDO (not LCM) DLLs, which were compiled against LibPalaso 2.6, for its older builds (builds that support data models 68, 69 and 70). Versions of Chorus that were built with the bugfix you mention were compiled against more recent LibPalaso versions, which include a change in namespaces from Palaso.Foo to SIL.Foo. I have been unable to get LfMerge to compile while linking against older FDO DLLs and newer Chorus DLLs: the IProgress impementations are incompatible between different LibPalaso versions. Which is why I have to use all the archived LibPalaso builds on TeamCity, and so on.

A lot of this pain will go away once I can drop support for FieldWorks 8 in the Language Forge Send/Receive process, but for now I have to use old Chorus DLLs in order to get something compiled against LibPalaso 6, and the Chorus bugfix was never backported to those old builds, which is why Language Forge Send/Receive has been failing for the past year, ever since FLEx 9 was released and people started upgrading to it.

rmunn (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/114)* status: 0 of 1 LGTMs obtained (waiting on @jasonleenaylor)

___
*[src/LfMerge.Core/Actions/SynchronizeAction.cs, line 98 at r1](https://reviewable.io/reviews/sillsdev/lfmerge/114#-M_fTAcP0RL2j-PGuDSl:-M_sHAlP02uj5cwz2jTu:b-8401zt) ([raw file](https://github.com/sillsdev/lfmerge/blob/2cb994c9eba407ab9e1bfefcefad7844784083cc/src/LfMerge.Core/Actions/SynchronizeAction.cs#L98)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

> ```
> 				try {
> 					if (!LfMergeBridge.LfMergeBridge.Execute("Language_Forge_Send_Receive", Progress,
> 						options, out syncResult))
> 					{
> 						Logger.Error(syncResult);
> 						return;
> 					}
> 				} catch (System.FormatException e) {
> 					if (e.StackTrace.Contains("System.Int32.Parse")) {
> 						ChorusHelper.SetModelVersion(MagicStrings.MinimalModelVersionForNewBranchFormat.ToString());
> 						return;
> 					} else {
> 						throw;
> 					}
> 				}
> ```
Braces are inconsistent, don't match the rest of the file, and also don't match general C# style guidelines. 

I'm not a developer on this tool so I can't speak with authority on this, but I would expect that you could fix the old version by just updating its version of chorus and avoid this exception entirely.
</blockquote></details>

Fixed in commit 8f37d9b. As for why I can't just update the version of Chorus, see the comment I left on the GitHub issue.

rmunn:
Commit 7e29616 is temporary, intended to use to run a Jenkins build and then back off the commit again once I see the build results. I'll do a force-push to revert to commit 8f37d9b once the Jenkins build runs.

rmunn:
Discovered that my bugfix so far isn't quite sufficient (see https://github.com/sillsdev/LfMerge/issues/107#issuecomment-850205243 for details). The new commits I just pushed should take care of that, where we'll now correctly extract DB version number 7000072 from a branch named `7500002.7000072`.

rmunn:
Note that #108 should be merged at the same time as this one; this one is against `fieldworks-master` and #108 is against `master`. Both branches need this bugfix.

rmunn:
Tested and working locally, and all Send/Receive unit tests passing. Merging. Once Jenkins builds the resulting packages, we can push them to QA and run them there for further testing.

PR #119, opened 2021-06-09
Fix custom fields bug

rmunn:
When custom fields referenced a ListReference and there was no GUID in the Mongo data, and the custom field had a wsSelector that was a negative number (a "magic" ws), Send/Receive would crash. This fixes that bug and includes a regression test to ensure the bug stays fixed.

This is commit f9d002d6abb4de8351a92bc7e924ff55e7614e1e from the master branch, rewritten for the fieldworks8-master branch.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/119)


rmunn:
Unit tests are passing locally, and this bugfix is already merged into master, so I'll merge it without waiting for a code review.

PR #120, opened 2021-06-21
Make iputils-ping a package, not dev, dependency

rmunn:
If ping is not present at runtime, Chorus fails to send anything. In some Docker deployments, the `iputils-ping` package is not installed by default, so we need `lfmerge` to depend in `iputils-ping` so that it will install correct dependencies in Docker/Kubernetes environments.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/120)


ermshiperete (approved):

megahirt:
Nice work, Robin.

PR #121, opened 2021-06-21
Make iputils-ping a package, not dev, dependency (FW8 branch)

rmunn:
If ping is not present at runtime, Chorus fails to send anything. In some Docker deployments, the `iputils-ping` package is not installed by default, so we need `lfmerge` to depend in `iputils-ping` so that it will install correct dependencies in Docker/Kubernetes environments.

This is https://github.com/sillsdev/LfMerge/pull/120 applied against `fieldworks8-master`. Both PRs should be merged together.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/121)


rmunn:
#120 was approved by @ermshiperete so I'm merging both.

PR #124, opened 2021-08-13
Update dependencies

ermshiperete:
This change updates the libpalaso* dependencies to the latest released versions.

Also move common properties to Directory.Build.props


rmunn:
This PR is not mergeable as-is, but should be left open so that we can examine the dependencies used in https://github.com/sillsdev/LfMerge/pull/278 and https://github.com/sillsdev/LfMerge/pull/284 and make sure that they were the latest ones. Once we've examined the dependency versions, we can close this PR without merging it, and possibly create a new PR with any needed dependency updates (which will be far simpler in terms of merge conflicts).

PR #126, opened 2021-08-18
Add LD_TRUST_TOKEN environment variable

rmunn:
Used for adding a trust_token query parameter to Mercurial URLs, which Language Depot will use to determine that the request came from the Language Forge server. (IP address checking doesn't work well with Kubernetes deployments, where the IP address can change frequently).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/126)

PR #127, opened 2021-08-20
Add LD_TRUST_TOKEN environment variable

rmunn:
Used for adding a trust_token query param to Mercurial URLs, which Language Depot will use to determine that the request came from the Language Forge server. (IP address checking doesn't work well with Kubernetes deployments, where the IP address can change frequently).

This is https://github.com/sillsdev/LfMerge/pull/126 targeted at `fieldworks8-master` rather than `master`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/127)


rmunn:
Merging without waiting for approval since #126 was already approved.

PR #129, opened 2021-08-23
Use LD_TRUST_TOKEN as Mercurial password

rmunn:
Mercurial doesn't accept query strings in repo URIs, so we have to submit the trust token as a password.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/129)


megahirt (approved):
LGTM

PR #138, opened 2021-09-20
Put env var name for trust token in MagicStrings

rmunn:
Now we have a single place to change it when we need to.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/138)


rmunn:
See also #139.

rmunn:

PR #139, opened 2021-09-20
Put env var name for trust token in MagicStrings

rmunn:
Now we have a single place to change it when we need to.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/139)


src/LfMerge.Core/MagicStrings.cs
```diff
@@ -94,6 +94,7 @@ static MagicStrings()
 
 		// Other magic strings that don't fall into any particular category
 		public const string UnknownString = "***";
+		public const string EnvVar_TrustToken = "LD_TRUST_TOKEN";
```
rmunn, line 97: Needs to become `LANGUAGE_DEPOT_TRUST_TOKEN`.

rmunn (commented):
Changes needed before this can be merged.

src/LfMerge.Core/MagicStrings.cs
megahirt, line 97: Per Robin's comment above:
```suggestion
		public const string EnvVar_TrustToken = "LANGUAGE_DEPOT_TRUST_TOKEN";
```

megahirt: I added a code suggestion to this effect

megahirt (approved):
LGTM.  Robin, what changes need to happen on K8s before we can deploy a new LFMerge with the new token name?

megahirt:
> LGTM. Robin, what changes need to happen on K8s before we can deploy a new LFMerge with the new token name?

Actually, now that I've merged master into this branch, it looks like no changes are made since the trust token name was already changed.  This PR was simply outdated.

megahirt (approved):
:lgtm:

> Reviewed 1 of 3 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/139#-:-N9tjx5u7GjD8La9RiRy:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained (waiting on @rmunn)

PR #142, opened 2021-11-16
Handle new "tags" field in Language Forge

rmunn:
Once https://github.com/sillsdev/web-languageforge/issues/1008 is implemented, LF data is going to start including a "tags" field, which LfMerge will need to handle.

This is a draft PR which I'll be working on at the same time as the implementation of the LF feature. They'll both need to be merged together, so this one will remain in draft status until both PRs are done.


rmunn:
This needs testing locally before it's actually ready to merge, but the bulk of the logic is there now so I'm un-marking this PR as draft. Any subsequent commits will be bug fixes, rather than major logic changes.

rmunn:
Needs testing and corresponding LF work; on hold until I am able to work on it again.

PR #146, opened 2021-11-26
Remove the need for monkeypatching after git clone

rmunn:
Now that the Docker build is the official build, we don't need to
preserve compatibility with the Jenkins build, so we'll remove the step
that monkeypatches the fieldworks8-master branch. This will simplify the
Dockerfile considerably.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/146)


rmunn:
Because of how GitHub Actions works for pull requests, this is running the GHA workflow from `fieldworks8-master` rather than the one from this PR, which means it's also running the Dockerfile from `fieldworks8-master`, and it fails on just exactly the step that this PR is intended to fix.

Which means we need to merge the PR despite the failing tests. Why does that remind me of "We have to pass the bill so you can find out what is in it"?

PR #153, opened 2021-11-27
No more checking out the Dockerfile in FW8 builds

rmunn:
The fieldworks8-master branch now needs its own version of the
Dockerfile, because it needs to have some code (adjusting the
AssemblyInfo.cs files) that's not needed in the master branch. Checking
out the master branch copy was causing PR 152 not to actually build what
I thought it was building.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/153)


rmunn:
Note that I don't want the fieldworks8-master and master branches to have separate Dockerfiles for long, as that's a path to long-term maintenance headaches. I will recombine them soon, but this is needed so that #152 will build correctly.

PR #154, opened 2021-11-28
Update AssemblyInfo files manually in FW8 builds

rmunn:
The GenerateAssemblyInfo feature came in with .Net Core 3.1. Anything built with Mono 5 won't have it, so for the FW8 builds we need to update AssemblyInfo.cs files ourselves. I haven't found a good GHA action for that, so we'll do it ourselves with a simple bash script. The bash script will reorder lines in the file, but that's not a big deal, and it keeps it very simple.

This PR ensures that the `fieldworks8-master` and `master` branches have the exact same build process, with the only differences controlled by conditionals and environment variables instead of having two different build workflows.

Eventually I plan to switch to having a single [reusable workflow](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) kept in the `master` branch, which will be *called* via `workflow_call` from all the other branches, to ensure that there's a Single Source of Truth™ for the build process.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/154)

### DIFF
--- a/.github/workflows/trythis.yml
+++ b/.github/workflows/trythis.yml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Get output version
         env:
-          VERSION: ${{ steps.id_version.outputs.version || steps.branch_version.outputs.version || steps.pr_version.outputs.version || 'no version found' }}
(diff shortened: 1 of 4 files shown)


rmunn:
This is basically #152 merged into `master` to ensure that the build files are the same across all the branches.

PR #157, opened 2021-11-30
Sign Debian packages when creating them

rmunn:
Note that we create the secrets folder as late as possible, and delete it as soon as possible, *even* if the build fails. Don't want anything to accidentally collect the secret key as a build artifact due to a mistyped glob pattern.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/157)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -152,17 +152,30 @@ jobs:
         if: steps.should_run.outcome == 'success'
         run: docker image ls
 
+      - name: Create and populate secrets folder
+        if: steps.should_run.outcome == 'success'
+        env:
+          SIGNING_KEY: ${{ secrets.PACKAGE_SIGNING_KEY_SIMPLE }}
+        run: |
+          mkdir -p secrets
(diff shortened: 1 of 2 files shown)


rmunn:
Just realized that this is unnecessary if we use GitHub Releases to distribute the packages, because then the source of trust that the packages are genuine will be the GitHub site and process. With Debian repos, which are often accessed over http, the repository needs to distribute a Release.gpg file that contains checksums of all the packages, which then gives confidence that no package has been corrupted or substituted because its SHA256 sum still matches the sum in the signed file. With a GitHub Release, the HTTPS certificate serves a similar function: it proves that you really are connecting to GitHub, and therefore the files in that Release are really the ones that the repo owner(s) chose to put in the Release. So as long as you trust GitHub's authorization system to work properly (the only people able to upload files to a Release are the people with write access to the repo), then you have sufficient confidence that the Release files came from the repo owner(s).

Therefore, I don't need to mess around with signing keys and passphrases here; I can just upload the built .deb files to a GitHub Release in the workflow and that will be sufficient.

Closing this PR as currently unnecessary.

PR #162, opened 2021-11-30
Tag fieldworks8-live builds, but do not release

rmunn:
* Tag fieldworks8-live builds, but do not release

Any time the fieldworks8-live build runs, we do need to push its Git tag
so that subsequent builds will pick up on the new tag

* Prevent potential race condition in tagging

If a fieldworks8-live build is kicked off and allowed to run to
completion before a live build is kicked off, the following could
happen:

- fw8-live build completes, tags v1.5.25 (for example)
- live build checks out fw8-live branch
- live build finds v1.5.25 tag, calculates new version number 1.5.26
- live build creates release with 1.5.26 packages in it

Later, another build is kicked off where fieldworks8-live and live
branches are pushed almost simultaneously, resulting in the following
build order:

- live build checks out fw8-live branch
- live build finds v1.5.25 tag, because v1.5.26 tag not yet in fw8-live
- live build creates release with 1.5.26 packages in it
- fw8-live build completes, tags v1.5.26

Result: two releases were created with lfmerge-fw8 packages at v1.5.26.

To avoid this potential issue, we only increment the patch number if
there has been at least one commit since the latest tag, otherwise if
we're building from a tagged commit then we rebuild the exact same
version number on that tagged commit.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/162)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -55,9 +55,12 @@ jobs:
           MINOR=$(echo "$DESCRIBE" | sed -E 's/^v([0-9]+)\.([0-9]+)\.([0-9]+).*$/\2/')
           PATCH=$(echo "$DESCRIBE" | sed -E 's/^v([0-9]+)\.([0-9]+)\.([0-9]+).*$/\3/')
           # TODO: Detect need for minor/major updates and increment those instead of PATCH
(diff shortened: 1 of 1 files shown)


rmunn:
This is #161, merged into `master` to keep both builds in sync.

PR #173, opened 2021-12-07

rmunn:

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/173)


megahirt (commented):
looks good so far!

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/173#-:-MqK_VyN1ncn9cSF-KUL:b-xo2sg4)* status: 0 of 1 LGTMs obtained (waiting on @rmunn)

rmunn:
Just ran a test, and got Send/Receive to complete correctly, but the custom field ended up with a type of "basic" instead of "multitext". Should be pretty straightforward to figure out why, but this isn't quite ready to merge yet.

rmunn:
Tested and working now, so now I can remove the bit from the Dockerfile that pulls in the web-languageforge image. That should make the lfmerge images much smaller and speed up the build time. Then it will be ready for review.

megahirt (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/173#-:-MqNz2I54IJAwy2Hejwv:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained (waiting on @rmunn)

PR #174, opened 2021-12-08

rmunn:

This is PR #173 but against `fieldworks8-master` rather than `master`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/174)


rmunn:
Whoops, didn't get the target branch correct. Closing, and will re-open with correct target branch.

PR #175, opened 2021-12-08

rmunn:

This is PR #173 but against fieldworks8-master rather than master.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/175)


megahirt (approved):
:lgtm:

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/175#-:-MqNykgdEnMOQisNuoXn:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained (waiting on @rmunn)

PR #191, opened 2022-02-11
Allow pbuild.sh and GHA workflow to both run

rmunn:
This should create the `builder` user with an appropriate uid in both the pbuild.sh and GHA workflow environments.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/191)

PR #192, opened 2022-02-11
Allow pbuild.sh and GHA workflow to both run

rmunn:
This should create the `builder` user with an appropriate uid in both the pbuild.sh and GHA workflow environments.

This is PR #191, against fieldworks8-master. I'll merge it along with #191 once #191 is approved.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/192)


rmunn:
#191 has been approved, so merging this one too.

PR #197, opened 2022-02-28
Get settings from environment variables

rmunn:
WIP. Once I finish changing all the places that reference LfMergeSettings (particularly in the test code that's pretty tightly coupled to the current implementation, which it mocks), LfMerge will get its settings from environment variables instead of an .ini file, allowing easier containerization.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/197)

### DIFF
--- a/src/LfMerge.Core/MongoConnector/MongoConnection.cs
+++ b/src/LfMerge.Core/MongoConnector/MongoConnection.cs
@@ -82,7 +82,7 @@ namespace LfMerge.Core.MongoConnector
 			var clientSettings = new MongoClientSettings();
 			// clientSettings.WriteConcern = WriteConcern.WMajority; // If increasing the wait queue size still doesn't help, try this as well
 			clientSettings.WaitQueueSize = 50000;
-			clientSettings.Server = new MongoServerAddress(Settings.MongoDbHostName, Settings.MongoDbPort);
+			clientSettings.Server = new MongoServerAddress(Settings.MongoHostname, Settings.MongoPort);
 			return new MongoClient(clientSettings);
 		}
 
--- a/LfMerge.TestApp/TestAppMainClass.cs
+++ b/LfMerge.TestApp/TestAppMainClass.cs
@@ -18,14 +18,11 @@ namespace LfMerge.TestApp
 				return;
 
 			var folder = Path.Combine(Path.GetTempPath(), "LfMerge.TestApp");
-			LfMergeSettings.ConfigDir = folder;
+			System.Environment.SetEnvironmentVariable(MagicStrings.SettingsEnvVar_BaseDir, folder);
 
 			MainClass.Container = MainClass.RegisterTypes().Build();
 			var settings = MainClass.Container.Resolve<LfMergeSettings>();
-			var config = new IniData();
-			var main = config.Global;
-			main["BaseDir"] = folder;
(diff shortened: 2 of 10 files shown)


rmunn:
Unit tests pass with the latest commit, so I'm declaring this ready for review.

src/LfMerge.Core/MongoConnector/MongoConnection.cs
```diff
@@ -82,7 +82,7 @@ private MongoClient GetNewConnection()
 			var clientSettings = new MongoClientSettings();
 			// clientSettings.WriteConcern = WriteConcern.WMajority; // If increasing the wait queue size still doesn't help, try this as well
 			clientSettings.WaitQueueSize = 50000;
-			clientSettings.Server = new MongoServerAddress(Settings.MongoDbHostName, Settings.MongoDbPort);
+			clientSettings.Server = new MongoServerAddress(Settings.MongoHostname, Settings.MongoPort);
```
megahirt, line 85: I like the property rename...I just can't see where this is defined.  I would have assumed that the rename in this usage would have required a rename in the definition of the property, but maybe I'm not understanding or seeing something.

megahirt (approved):
LGTM - thank you Robin.  Some good refactoring and code deletion in this PR as well.  I have one non-blocking comment/question.

src/LfMerge.Core/MongoConnector/MongoConnection.cs
rmunn: It used to be defined here: https://github.com/sillsdev/LfMerge/blob/b44d9eb60a579cfa9b553d20f8c345c5a149eb8c/src/LfMerge.Core/Settings/LfMergeSettings.cs#L92

The changes to LfMergeSettings.cs were so extensive that GitHub ended up hiding the diff by default, and even if you click on "View diff" the side-by-side diff rendering ends up being a little useless, as it's matching up lines that don't really match, such as opening curly braces from different functions, because those are the closest non-changing lines it can find. So it's probably best to review LfMergeSettings.cs by opening the old and new versions in different browser tabs and switching between them.

Old version of LfMergeSettings.cs: https://github.com/sillsdev/LfMerge/blob/b44d9eb60a579cfa9b553d20f8c345c5a149eb8c/src/LfMerge.Core/Settings/LfMergeSettings.cs
New version of LfMergeSettings.cs: https://github.com/sillsdev/LfMerge/blob/feature/settings-from-env-vars/src/LfMerge.Core/Settings/LfMergeSettings.cs

rmunn:
@megahirt When you have time, you might want to take a look at this again. When you highlighted the change from MongoDbHostName to MongoHostname, you mentioned that you didn't see where that was defined. That means that you may not have seen the very large diff that GitHub hid by default, which was the core of this PR. So in case your review was based on incomplete information, I wanted to bring https://github.com/sillsdev/LfMerge/pull/197#discussion_r826737490 to your attention.

src/LfMerge.Core/MongoConnector/MongoConnection.cs
megahirt: You're right 🤣 I missed the core changes of this PR due to the diff being collapsed.  Oh well.  LGTM.

megahirt:
I approve, ready to merge

PR #204, opened 2022-04-06
Tag LfMerge branches when release is pushed (FW8)

rmunn:
GHA workflow will now tag the repo with same version number being put into the Docker image tags.

We also separate out the `release` workflow into its own file, as a first step towards someday building a truly reusable Docker-release workflow. (It's not there yet; this version of the workflow has some LfMerge assumptions, like branch structure, baked into it).

This will only run when the `live` branch is pushed, but I'm putting this PR into both the `master` and `fieldworks8-master` branches so that they'll continue to have identical workflows.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/204)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -150,6 +150,14 @@ jobs:
           echo "::set-output name=AssemblySemFileVer::${AssemblySemFileVer}"
           echo "::set-output name=InformationalVersion::${InformationalVersion}"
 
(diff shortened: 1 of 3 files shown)


rmunn:
[Thoroughly tested in my test repo](https://github.com/rmunn/Testing/actions/runs/2101268484) before making this PR, so not waiting for a review.

PR #206, opened 2022-04-08
Hopefully speed up Docker image pull

rmunn:
This should hopefully speed up pulling Docker images for lfmerge by making the "apt-get" steps more consistent and more likely to have cache hits, thereby making the big image layers more likely to be in common between different builds.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/206)


rmunn:
All checks passed, so I'm going to squash and merge without review since I know it works.

PR #236, opened 2022-05-18
Slimmer LfMerge images

rmunn:
We now build from a base image stored in the GitHub Container Registry.
By storing the base image in GHCR and only updating it when once of the
lfmerge dependencies changes (which almost never happens), we can avoid
re-running `apt-get update` all the time, which always ends up causing a
layer change and the entire 700MB layer to be different. This results in
much slimmer lfmerge images and faster download times.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/236)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,7 @@
 # syntax=docker/dockerfile:experimental
 ARG DbVersion=7000072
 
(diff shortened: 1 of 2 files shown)


rmunn:
All builds succeeded, but #237 build failed due to an HTTP timeout. Re-running the build in #237, then if it passes I'll merge both of these at once.

PR #240, opened 2022-05-20
Get unit tests running again in Docker containers

rmunn:
Unit tests were failing after switching the LfMerge build to a Dockerized environment. This PR gets them working again, at least on the FW 9 branch. (A separate PR will be needed for the FW 8 branch, of course).

The main change we had to account for was the change to LfMergeSettings and how it now takes environment variables as settings. The LfMergeSettings test doubles now set the vars they need before calling SetAllMembers().

**NOTE:** This PR does **not** run unit tests by default in the build, because they take too long (20+ minutes) to be useful. Instead, I intend to create another PR that divides the tests up into fast and slow tests, then runs the fast tests with every PR and allows us to run the slow tests on-demand.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/240)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 data/
 !data/testlangproj
+!data/testlangproj-modified
 finalresults*
 Dockerfile
 packages/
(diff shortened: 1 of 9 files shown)


rmunn:
This PR's logic was incorporated into https://github.com/sillsdev/LfMerge/pull/284, which has been merged and released. It's no longer needed.

PR #241, opened 2022-05-25
Fix build of FW 8 versions

rmunn:
The FW 8 versions no longer build into output/Release/net462, but only
into output/Release. So we need to grab the build output from there.
I've kept the ability to get build output from output/Release/net462 as
well, just in case this configuration changes again in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/241)

### DIFF
--- a/.github/workflows/docker-build-fw9only.yml
+++ b/.github/workflows/docker-build-fw9only.yml
@@ -185,7 +185,7 @@ jobs:
 
       - name: Collect tarball images
       # Now should collect tarballs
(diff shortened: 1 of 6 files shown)


rmunn:
Build won't succeed until both this PR and #242 are merged. I've tested this locally and it works, so I'm going to merge this even though the build is failing. Because #242 is succeeding on all DbVersions except 72, while the FW9-only build succeeds on DbVersion 72 in this PR. So all is well.

PR #246, opened 2022-05-26
Tweak build

rmunn:
This PR is for running the build over and over and seeing what happens.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/246)


rmunn:
Finally found the real issue, and opened #248 and #249 to deal with it. So this PR can be closed now.

PR #248, opened 2022-05-26
Fix DbVersion in FW9 build

rmunn:
DbVersion was always being set to 7000072, because I wasn't defining the DatabaseVersion property in the msbuild invocation and the .csproj files were always defaulting to 7000072. This was correct for the FW 9 build, but the FW 8 build was always defaulting to 7000070 and that was wrong for the 68 and 69 builds. Now set correctly throughout.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/248)


rmunn:
#249 has the biggest fix, but this PR is necessary so that the FW 8 and 9 builds don't diverge.

PR #269, opened 2022-06-02
Do not put projects on hold for DNS errors

rmunn:
DNS errors are likely transient, so don't put projects on hold if that happens.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/269)


rmunn:

PR #273, opened 2022-06-10
Exit LfMerge container faster by handling SIGTERM

rmunn:
This will allow the container to shut down cleanly in docker-compose and Kubernetes, rather than waiting for 10 seconds when it doesn't answer SIGTERM and then eventually being sent SIGKILL.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/273)

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -9,6 +9,10 @@ trap "exit" TERM
 # Ensure /var/log/syslog exists so tail -f /var/log/syslog will run
 logger "Starting container..."
 # echo /var/log/syslog to container stdout so it shows up in `kubectl logs`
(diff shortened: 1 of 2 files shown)


rmunn:
While I was working on the entrypoint script, I also fixed the issue I'd observed in testing where sometimes, `tail -f /var/log/syslog` would run before the syslog file had actually been created, resulting in LfMerge logs not being visible on stdout with `docker logs -f lfmerge` or `kubectl logs -f deploy/lfmerge`. By waiting a short period of time for syslog to exist, we can ensure that the lfmerge container logs will always be visible.

Issue #232, opened 2022-05-12
Catch all exceptions when iterating over entries and comments, both ways.  Skip and log the problem at the entry level, then continue with the sync


megahirt:
This task is all about making LFMerge more robust by reducing the number of projects on HOLD.  Any exception that occurs while in the entry or comment loop within either TransferMongoToLcm or TransferLcmToMongo should be handled by

- log the exception error to the state file
- skip the entry or comment and carry on with the next

Additional information about exception handling should now be kept the sync state file.  I propose the following changes to the state file:

Remove "ErrorMessage" in favor of an "Error" object with multiple properties.
Behavior:
- If no exceptions are thrown in a LfMerge sync, the Error object is discarded and the state is set to IDLE
- If an exception is thrown in a place other than the looping over entries / comments, the exceptionMessage is filled out and the state changes to HOLD.
- LFMerge will catch all exceptions inside the loops over entries and comments and log the error messages, as well as metadata about the comment/entry which failed to sync.  Skip the failed entry and increment the skip counter.  The sync will finish with a log of errors available to the user, and the state is set back to IDLE

- the proposed Error object looks like this ([also attached as a json file for easy copy/paste](https://github.com/sillsdev/LfMerge/files/8674141/lfmerge.error.sample.json.txt)):
```
{
  "Error": {
    "timestamp": "timestamp of when the exception occurred (or in the case of skipped entries/comments) the time when processing finished",
    "exceptionMessage": "string of global error message that occurred (not while looping entries/comments)",
    "skipped": {
      "toMongo": {
        "entries": {
          "count": 5,
          "list": [
            {
              "guid": "FW GUID",
              "id": "MongoId",
              "exceptionMessage": "message",
              "label": "string representing the entry - probably lexeme first non empty WS. 100 character limit"
            }
          ]
        },
        "comments": {
          "count": 3,
          "list": [
            {
              "guid": "FW GUID",
              "id": "MongoId",
              "exceptionMessage": "message",
              "label": "string representing the comment- 100 char limit",
              "entry": {
                "guid": "FW GUID",
                "id": "MongoId",
                "label": "string representing the comment- 100 char limit"
              }
            }
          ]
        }
      },
      "toLcm": {
        "entries": {
          "count": 5,
          "list": [
            {
              "guid": "FW GUID",
              "id": "MongoId",
              "exceptionMessage": "message",
              "label": "string representing the entry - probably lexeme first non empty WS. 100 character limit"
            }
          ]
        },
[...]

PR #278, opened 2022-06-24
Updated LfMerge to .NET6

josephmyers:
Includes updating .Core to .NET Standard. I picked this simply to instill more compatibility with other frameworks. Also includes updating various old Framework packages to the minimum versions that supported .NET Standard.
Removed icu.net to depend on indirect reference. Upgraded MongoDB, per https://github.com/commandlineparser/commandline/issues/225 and refactored some duplicated code into a base class in .Core. Referencing the "netcore" versions of the SIL packages -- these can go back once the "netcore" branches are in. Upgraded libpalaso references to 9, since most are already .NET Standard.

There are lots of warnings for .NET Framework packages, some fixable, like SIL.Lexicon, and some not, like NHunspell. But these changes allow it to dotnet build

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/278)


rmunn:
Looking forward to seeing this completed; building and running with .NET 6, if you can manage to get it all working, will allow me to use several features from recent C# versions that I've been looking forward to having available. It'll make for much nicer LfMerge code in the long run. Thanks for working on this!

josephmyers:
@rmunn so the build is passing now, and i think i've gotten the usages of mono removed. however, i am definitely not the expert on the innerworkings of the surrounding scripts, so there may be some (major) things i'm missing. but feel free to take a preliminary look!

note that the GHA workflows were flaky -- i had to do manual re-runs to get the FW8 ones to pass. Assuming this is not normal

rmunn:
Did some testing, and tweaked the build until it works. It fails to run, because the SyslogLogger class is looking for the Mono.Posix.NETStandard assemblty, which of course doesn't exist in a non-Mono runtime environment. But that will be easy to fix: just swich LfMerge over to logging everything to stdout, which is better in a Docker environment anyway, and get rid of the SyslogLogger. I'll work on that with you tomorrow.

rmunn:
I got the missing DLL installing with the rest of the LfMerge files, but .NET still thinks it's missing, and I don't know why yet. But I did figure out where its dependency is coming from: it's depended on by the SIL.WritingSystems package, which is a core part of LfMerge that we can't do without. We call `Sldr.Initialize()` during our setup, which tries to acquire a filesystem lock. It uses SIL.Threading.GlobalMutex.Initialize() to do so, and SIL.Threading.GlobalMutex uses SIL.Threading.GlobalMutex.LinuxGlobalMutexAdapter.Init which in turn looks for the Mono.Posix.NETStandard assembly.

So we need to look at the SIL.Threading package and see if that can be converted to .NET Standard so as to not be looking for a Mono DLL.

josephmyers:
Mono.Posix.NETStandard is used by SIL.Core and SIL.LCModel.Utils, both of which are .NET Standard. if we need to remove these, I can look into that

rmunn:
Looks like Mono.Posix is the older name, and the name has been changed to Mono.Unix: https://www.nuget.org/packages/Mono.Unix/7.1.0-final.1.21458.1

It might be necessary to update the .NET Standard versions of libpalaso to use Mono.Unix, or it might not. We'll find out, most likely.

rmunn:
According to [this issue comment](https://github.com/L3tum/HardwareInformation/issues/15#issuecomment-573791709), the Mono.Posix package may not work in .NET 4.8, let alone .NET 6.0. We'll definitely want to look into moving to the Mono.Unix packages for this.

rmunn:
I've opened https://github.com/sillsdev/libpalaso/issues/1186 to track the necessary work in libpalaso for switching to Mono.Unix.

josephmyers:
> According to [this issue comment](https://github.com/L3tum/HardwareInformation/issues/15#issuecomment-573791709), the Mono.Posix package may not work in .NET 4.8, let alone .NET 6.0. We'll definitely want to look into moving to the Mono.Unix packages for this.

This is interesting. In that comment he mentions the theoretical .NET Standard package being compatible, so I'm not fully convinced there's something inherently wrong with all Mono.Posix.NetStandard packages. I think we'll just have to make a local package and see if it works

josephmyers:
So this morning I verified Mono.Posix.NETStandard works in a .NET6 environment, using the following simple program:
[Consuming.zip](https://github.com/sillsdev/LfMerge/files/9051554/Consuming.zip)

Now I'm not saying we shouldn't move away from it. I'm just saying there's something else going on with that exception

josephmyers:
@rmunn have you tried `dotnet build` from the terminal? if i use that instead of `pbuild.sh`, i get past the FileNotFoundException. so there's some difference in how `pbuild` is setting things up and `dotnet build`

josephmyers:
@rmunn a bit more progress on this, and maybe you can provide further guidance. i noticed the `Mono.Posix.NETStandard` structure was different between `dotnet build` and the `'pbuild.sh`, so i copied the `runtimes/` folder from `dotnet build` into the `pbuild.sh` version, and it got past the error.

`runtimes/` has multiple copies of `Mono.Posix.NETStandard`, as well as a few other lib's, based on OS/architecture. the one we needed is from `linux-x64`. so how do we update the pbuild process to duplicate this `runtimes/` structure? note this folder structure was required -- pasting the lib into the folder with the executable didn't help.

rmunn:
Haven't managed to look into this in-depth, but a quick look showed the lfmerge process running successfully (yay), but displaying the error message "Can't acquire file lock - is another instance running?" on first run (in a clean container where there is no lock file yet). This error message is only displayed when the SIL.IO.FileLock.SimpleFileLock class returns false from the TryAcquireLock() method. If there was no other lfmerge instance running, then failing to acquire the lock is usually indicative of some sort of failure in the Mono.Posix or Mono.Unix calls that are used to acquire the file lock.

In other words, the "Can't acquire file lock" error probably means that there's still more work to do before this is working, alas.

josephmyers:
The error is from SimpleFileLock in SIL.Core. File.Create() for /var/run/ is throwing the exception with this new framework. This is probably a simple fix, but I'm not good enough with linux operations in C# yet to have figured it out

rmunn:
Commit abe050f fixes the permissions issue.

josephmyers:
image

@rmunn  here's the exception i was seeing yesterday. obviously if that directory already exists, maybe from a .NET Framework run, no problem. i'm guessing my computer is a slightly cleaner test environment than yours

josephmyers:
> 
> @rmunn here's the exception i was seeing yesterday. obviously if that directory already exists, maybe from a .NET Framework run, no problem. i'm guessing my computer is a slightly cleaner test environment than yours

Well you may be able to disregard this. with that path missing, running `make` doesn't throw that exception, and i'm assuming that's a good thing

megahirt (approved):
There are more files in here than I can reasonably look over but those I have glanced at, I am happy to put my stamp of approval on.  Knowing that Robin has already reviewed this makes this mostly a rubber stamp.

rmunn (approved):
LGTM. Tested locally, and it successfully builds in GitHub Actions, so it's time to merge this into `master` and test on the staging server. I'm sure we'll find some other tweak that needs to be made once we test in a live environment, but this is ready to merge. Thanks for the good work, @josephmyers!

rmunn:
We'll deploy this when #232 is deployed.

PR #280, opened 2022-07-14
Skip failing entries in sync, rather than putting projects on hold

### LINKED ISSUE #232: Catch all exceptions when iterating over entries and comments, both ways.  Skip and log the problem at the entry level, then continue with the sync

rmunn:
Fixes #232. Will need corresponding changes to Language Forge to display the failing entries (if any) to the user; see https://github.com/sillsdev/LfMerge/issues/232#issue-1233306693 for the proposed JSON structure of the new error object. Note that the JSON property will change from `errors` to `error` (it was `Error` in the https://github.com/sillsdev/LfMerge/issues/232#issue-1233306693 proposal, but it's going to be `error` once this PR is merged).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/280)

### DIFF
--- a/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
+++ b/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
@@ -40,6 +40,7 @@ namespace LfMerge.Core.DataConverters
 			LfProjectConfig config = _factory.Create(_project).Config;
 			FieldLists fieldConfigs = FieldListsForEntryAndSensesAndExamples(config);
 
+			var exceptions = new List<CommentConversionError<ILexEntry>>();
 			var fixedComments = new List<LfComment>(_conn.GetComments(_project));
 			string allCommentsJson = JsonConvert.SerializeObject(fixedComments);
 			// _logger.Debug("Doing Lcm->Mongo direction. The json for ALL comments from Mongo would be: {0}", allCommentsJson);
@@ -55,7 +56,6 @@ namespace LfMerge.Core.DataConverters
 				List<KeyValuePair<string, Tuple<string, string>>> statusChanges = JsonConvert.DeserializeObject<List<KeyValuePair<string, Tuple<string, string>>>>(newStatusChangesStr);
 
 				var entryErrorsByGuid = _entryConversionErrors.EntryErrors.ToDictionary(s => s.EntryGuid());
-				var exceptions = new List<CommentConversionError<ILexEntry>>();
 				foreach (LfComment comment in comments)
 				{
 					// LfMergeBridge only sets the Guid in comment.Regarding, and leaves it to the LfMerge side to set the rest of the fields meaningfully
--- a/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
+++ b/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
@@ -37,8 +37,6 @@ namespace LfMerge.Core.DataConverters
 
 		public List<CommentConversionError<ILexEntry>> RunConversion()
 		{
-			var entryErrorsByGuid = _entryConversionErrors.EntryErrors.ToDictionary(s => s.EntryGuid());
-			var exceptions = new List<CommentConversionError<ILexEntry>>();
 			LfProjectConfig config = _factory.Create(_project).Config;
 			FieldLists fieldConfigs = FieldListsForEntryAndSensesAndExamples(config);
 
@@ -56,6 +54,8 @@ namespace LfMerge.Core.DataConverters
 				List<Tuple<string, List<LfCommentReply>>> replies = JsonConvert.DeserializeObject<List<Tuple<string, List<LfCommentReply>>>>(newRepliesStr);
 				List<KeyValuePair<string, Tuple<string, string>>> statusChanges = JsonConvert.DeserializeObject<List<KeyValuePair<string, Tuple<string, string>>>>(newStatusChangesStr);
 
+				var entryErrorsByGuid = _entryConversionErrors.EntryErrors.ToDictionary(s => s.EntryGuid());
+				var exceptions = new List<CommentConversionError<ILexEntry>>();
 				foreach (LfComment comment in comments)
 				{
 					// LfMergeBridge only sets the Guid in comment.Regarding, and leaves it to the LfMerge side to set the rest of the fields meaningfully
@@ -95,9 +95,12 @@ namespace LfMerge.Core.DataConverters
 				}
 				var skippedCommentGuids = new HashSet<Guid>(exceptions.Select(s => s.CommentGuid()));
 				var skippedCommentGuidStrs = new HashSet<string>(skippedCommentGuids.Select(s => s.ToString()));
-				_conn.UpdateComments(_project, comments.Where(s => !(s.Guid.HasValue && skippedCommentGuids.Contains(s.Guid.Value))).ToList());
-				_conn.UpdateReplies(_project, replies.Where(s => !skippedCommentGuidStrs.Contains(s.Item1)).ToList());
-				_conn.UpdateCommentStatuses(_project, statusChanges.Where(s => !skippedCommentGuidStrs.Contains(s.Key)).ToList());
+				var unskippedComments = comments.Where(s => !(s.Guid.HasValue && skippedCommentGuids.Contains(s.Guid.Value)));
+				var unskippedReplies = replies.Where(s => !skippedCommentGuidStrs.Contains(s.Item1));
+				var unskippedStatusChanges = statusChanges.Where(s => !skippedCommentGuidStrs.Contains(s.Key));
+				_conn.UpdateComments(_project, unskippedComments.ToList());
+				_conn.UpdateReplies(_project, unskippedReplies.ToList());
+				_conn.UpdateCommentStatuses(_project, unskippedStatusChanges.ToList());
 			}
 			else
 			{
--- a/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
+++ b/src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
@@ -8,6 +8,7 @@ using LfMerge.Core.Logging;
 using LfMerge.Core.MongoConnector;
 using LfMerge.Core.LanguageForge.Config;
 using LfMerge.Core.LanguageForge.Model;
+using LfMerge.Core.Reporting;
 using Newtonsoft.Json;
 using SIL.LCModel;
 using SIL.Progress;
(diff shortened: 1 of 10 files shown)


src/LfMerge.Core/ProcessingState.cs
```diff
@@ -109,6 +112,40 @@ public void PutOnHold(string errorMessage, params object[] args)
...
 		}
 
+		public void ReportMongoToLcmErrors(ConversionErrors toLcmErrors, Exception overallException = null)
+		{
+			if (_error != null && _error.Skipped != null) {
+				_error.Skipped.ToLcm = toLcmErrors;
+				_error.Timestamp = DateTime.UtcNow;
+			} else {
+				var skipped = ConversionErrorReport.Create(null, toLcmErrors);
+				if (_error != null) {
```
josephmyers, line 122: For these two methods, if the _error has already been set, why don't we use the overallException? This will mean the error report will have whatever the previous ExceptionMessage was. We should fix this

src/LfMerge.Core/Reporting/ErrorReport.cs
```diff
@@ -0,0 +1,162 @@
...
+using LfMerge.Core.LanguageForge.Model;
+using SIL.LCModel;
+
+namespace LfMerge.Core.Reporting
+{
+    public class ErrorReport
+    {
+        public DateTime Timestamp { get; set; }
+
+        public string ExceptionMessage { get; set; }
```
josephmyers, line 16: Unused -- recommend deletion

src/LfMerge.Core/Reporting/ErrorReport.cs
```diff
@@ -0,0 +1,162 @@
...
+                Timestamp = DateTime.UtcNow,
+                ExceptionMessage = overallException?.ToString(),
+                Skipped = skipped
+            };
+        }
+
+        public static ErrorReport Create(ConversionErrorReport skipped)
+        {
+            return ErrorReport.Create(skipped, null);
+        }
```
josephmyers, line 32: Since we have these nice creation methods, it will be ideal to make the property setters private. This will prevent future issues/bugs, like the one I point out in ProcessingState.

src/LfMerge.Core/Reporting/ErrorReport.cs
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) 2022 SIL International
+// This software is licensed under the MIT license (http://opensource.org/licenses/MIT)
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using LfMerge.Core.LanguageForge.Model;
+using SIL.LCModel;
+
+namespace LfMerge.Core.Reporting
```
josephmyers, line 10: Assuming the intent of these classes is to be used only in constructing the ErrorReport, I would make them all internal, except for maybe that one.

src/LfMerge.Core/Reporting/ConversionError.cs
```diff
@@ -0,0 +1,154 @@
...
+// This software is licensed under the MIT license (http://opensource.org/licenses/MIT)
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using LfMerge.Core.LanguageForge.Model;
+using SIL.LCModel;
+using LfMerge.Core.DataConverters;
+
+namespace LfMerge.Core.Reporting
```
josephmyers, line 11: These classes can all be internal, no?

josephmyers (dismissed):
Overall, you really cleaned up the code a lot from what we originally started with, so good job! The comments below should all be minor

src/LfMerge.Core/DataConverters/ConvertLcmToMongoComments.cs
```diff
@@ -18,22 +19,26 @@ public class ConvertLcmToMongoComments
...
 			_servLoc = proj.FieldWorksProject.ServiceLocator;
 			_logger = logger;
 			_progress = progress;
 			_factory = factory;
 		}
 
-		public void RunConversion()
+		public List<CommentConversionError<ILexEntry>> RunConversion()
 		{
+			var entryErrorsByGuid = _entryConversionErrors.EntryErrors.ToDictionary(s => s.EntryGuid());
```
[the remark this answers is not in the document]
rmunn: The catch block is inside a foreach and I don't want to create this dictionary every time through the loop, but I'll move it just before the foreach loop so it's closer to where it's used.

src/LfMerge.Core/DataConverters/ConvertLcmToMongoLexicon.cs
```diff
@@ -129,17 +131,34 @@ public void RunConversion()
...
+					LfLexEntry lfEntry = LcmLexEntryToLfLexEntry(LcmEntry);
+					lfEntry.IsDeleted = false;
+					Logger.Info("{3} - LcmToMongo: {0} LfEntry {1} ({2})", createdEntry ? "Created" : "Modified", lfEntry.Guid, ConvertUtilities.EntryNameForDebugging(lfEntry), i);
+					Connection.UpdateRecord(LfProject, lfEntry);
+				}
+				catch (Exception e)
+				{
+					exceptions.AddEntryError(LcmEntry, e);
+
+					//todo check if all entries have been failures
```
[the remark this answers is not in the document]
rmunn: We haven't done this yet; that's something I should either implement or decide not to do. I'll think about it.

src/LfMerge.Core/DataConverters/ConvertMongoToLcmComments.cs
```diff
@@ -17,46 +18,73 @@ public class ConvertMongoToLcmComments
...
+					} else {
+						exceptions.Add(new CommentConversionError<LfLexEntry>(comment, e, null));
+					}
 				}
-
-				commentsWithIds.Add(new KeyValuePair<string, LfComment>(comment.Id.ToString(), comment));
 			}
-			string allCommentsJson = JsonConvert.SerializeObject(commentsWithIds);
+			var skippedCommentGuids = new HashSet<string>(exceptions.Select(s => s.CommentGuid().ToString()));
+			string allCommentsJson = JsonConvert.SerializeObject(commentsWithIds.Where(s => !skippedCommentGuids.Contains(s.Key)));
```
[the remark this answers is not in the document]
rmunn: To match the names I used in ConvertLcmToMongoComments, I went with `unskippedCommentsJson`.

src/LfMerge.Core/Reporting/ErrorReport.cs
rmunn: Not unused; it's populated in the Create() method, which is called from the ProcessingState.Report(x)To(y)Errors methods.

src/LfMerge.Core/Reporting/ErrorReport.cs
rmunn: Things is, in some cases (like what we do in ProcessingState) we explicitly update the existing _error object instead of creating a new one. The reason for that is because the MongoToLcm and LcmToMongo parts of the error report are created at different times, so they need to populate the existing error report if it exists. Alternately, I could treat these as read-only classes (called "records" in the new C# feature set) and copy over the half that I'm not populating. If I redesign this later, I'll probably do that; but for now, making the property setters private would prevent the current code approach from working and would force a redesign.

src/LfMerge.Core/Reporting/ErrorReport.cs
rmunn: Wait, *internal*, not private. Never mind. Yes, those can probably be made internal, as I don't think they're referenced from other assemblies. Though I also don't think there's any particular reason to make them internal; I tend to default to openness in my programming style, rather than defaulting to locking the code down which seems to be the approach taken by most OOP style guides.

src/LfMerge.Core/Reporting/ConversionError.cs
rmunn: Yes, they could, but I'm not persuaded there would be any benefit to it. As I said in my previous comment, I don't really believe in locking everything down the way most OOP style guides teach. I believe in leaving things open unless they *need* to be locked down, and this doesn't, AFAICT.

src/LfMerge.Core/ProcessingState.cs
josephmyers: Fixed -- thanks!

src/LfMerge.Core/Reporting/ErrorReport.cs
josephmyers: With the above change, this is now used -- thanks

src/LfMerge.Core/DataConverters/ConvertLcmToMongoLexicon.cs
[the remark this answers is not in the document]
josephmyers: Deleted -- resolving

src/LfMerge.Core/Reporting/ErrorReport.cs
josephmyers: Yeah, I agree with OOP design, but you don't have to follow if you don't want to. Just a recommendation!

src/LfMerge.Core/Reporting/ConversionError.cs
josephmyers: The more open the structure is, the more prone it is to bugs, as we're seeing today. I and most people prefer clean, controlled API's, but I'm not going to push for this super hard. Feel free to resolve these two

rmunn:
I've addressed the concerns in Joseph's review. I'm going to dismiss his review since he's on vacation until the end of the week, and instead ask Chris to review this.

megahirt (approved):
I've looked over the code from my own 30,000ft view and see any issues.  Let's talk how we can test this change...

src/LfMerge.Core/Reporting/ErrorReport.cs
rmunn: Did the redesign in https://github.com/sillsdev/LfMerge/pull/280/commits/c9cca21bb436fce27e91d60fff22f7459334caf1, so I'll resolve this now.

PR #286, opened 2022-08-22
Added launch config for debugging in VS Code.

josephmyers:
This involved changing the "build" task, which may break other things. This change is a little brittle and thus may not be desirable longterm. I'm certainly open to suggestions. It builds LfMergeQueueManager (specifically) and puts the output in the pbuild.sh output directory. Launching from here uses the full-featured environment already set up for us by the build script.

This is low priority.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/286)


megahirt:
Anything like this that will make your job easier as a developer is a welcome change.  It would be great if there were simple instructions in the README or at least in this PR for how to drive the new task as part of the larger task of "develop and build LFMerge".  The more documentation we spread here and there, the better.

megahirt (approved):

megahirt:
I can confirm that the launch configuration works on my machine - thanks @josephmyers for this contribution!

PR #287, opened 2022-08-23
baseline for working unit tests via pbuild

josephmyers:
improvements for finding ChorusMerge, depending on platform
moved LfMerge.Core from .NET Standard to .NET6, to fix warnings about consuming ChorusMerge
this change won't work without a local package of ChorusMerge. it needs the updates to Program.cs. once that PR is merged, we can update the reference here and remove the need for a local nuget source

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/lfmerge/287)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ MercurialExtensions/
 Mercurial/
 Downloads/
 lib/
+!lib/chorusmerge
 build/LfMerge.files
 data/php/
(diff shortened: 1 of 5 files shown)


rmunn:
This has been incorporated into #290, but I'd like to keep it open until #290 has been updated with the actual Chorus build. Once that's been proven to work, then we can merge #290 and close this one.

PR #288, opened 2022-08-24
Rename `master` to `develop`, drop `qa` branch

rmunn:
Also drop GitVersion.yml since we no longer use or need it.

Next step will be to rename `live` to `master`, but we'll do that *after* the `master` to `develop` rename has gone through, so that anyone with local copies of the repo doesn't find themselves on the branch formerly known as `live` when they should hvae been on `develop`.


rmunn:
After this PR merges, we'll rename the `master` branch to `develop` in the GitHub UI.

josephmyers (changes requested):
Code looks good. Don't forget to update the readme

josephmyers (approved):

PR #290, opened 2022-08-24
Get build working with fixed chorusmerge

rmunn:
This PR will eventually supersede #287. It requires downloading the build artifacts from https://github.com/sillsdev/chorus/pull/279 and unzipping them into a folder named `local` at the root of the LfMerge repository. Then the `dotnet restore` step will consume those packages (thanks to the `NuGet.config` file included in this PR), which means it will pull in the `chorusmerge` bugfix from https://github.com/sillsdev/chorus/pull/279. With that bugfix, all unit tests finally pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ MercurialExtensions/
 Mercurial/
 Downloads/
 lib/
+!lib/chorusmerge
 build/LfMerge.files
 data/php/
 data/testlangproj*/*.bak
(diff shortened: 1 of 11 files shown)


rmunn:
All but two unit tests passed on my machine when I ran `pbuild.sh` with unit tests enabled, and those two tests were brittle tests whose failure caused by a too-specific expectation (see log message of commit f4f915c for all the details). With commit f4f915c the tests should now pass, so this is ready for review.

josephmyers (approved):

PR #291, opened 2022-08-25
mkdir ~/.nuget/packages dir if necessary

megahirt:
This directory is expected on the host system and referenced in the pbuild.sh script.

Since I am a new C# dev, I did not have the directory already and the LFMerge build failed.  This ensures the build will succeed on brand new machines like mine.


rmunn (approved):
Good idea.

PR #292, opened 2022-08-26
Run unit tests in GHA workflow and report results

rmunn:
This runs the unit tests by default in the GHA workflow (setting RUN_UNIT_TESTS to 1 by default) and reports the test results in the GHA workflow logs. It uses the https://github.com/marketplace/actions/publish-test-results action to report the results; see that documentation for how to configure the reporting locations if desired.


rmunn:
Improvements that could be made:

* Add some build args to the Dockerfile to specify whether the tests are running in a PR or on the develop or master branches
* Apply a filter to NOT run the tests in the "IntegrationTests" category on every PR, to save time while iterating... but *do* run them manually (in a different workflow?) before merging a PR. And *do* run them on develop and master branches, of course.

rmunn:
Okay, this won't work. The tests all *passed*, but the dotnet test command failed with the error "The active test run was aborted. Reason: Test host process crashed". This is apparently a bug that's been around since at least July 2021 in the `dotnet test` command, but it means that we can't use the return code of the `dotnet test` process to distinguish between a passing test suite and a test suite with one or more failures. So this PR won't work until that `dotnet test` bug gets fixed.

rmunn:
With commit https://github.com/sillsdev/LfMerge/pull/292/commits/a69169b78b54e92f3241a6691bc4a621c5971a04 this works now, but we still need to add a step that will check the test results and refuse to create a release if the test results were a failure. That's something I'm running out of time to implement, so it will have to wait for a later PR.

megahirt (approved):
Great improvements!

PR #304, opened 2022-09-01
Removed QueueNames

josephmyers:
This included removing IQueue.Name and IQueue.NextQueueWithWork as well as simplifying container registration for Queue.

Also moved queue directory creation from Queue to LfMergeSettings.


longrunningprocess (approved):
this looks so much better to me... since I don't have any cs experience though maybe you'd want to wait to get someone with a better sense of conventions and approaches.  If the tests all pass however, you have my blessing fwiw!

megahirt (approved):
LGTM - let's not ship this to master until we've shipped and tested the many other changes in LFMerge in production

PR #306, opened 2022-09-20
LfMerge completes S/R's with .NET6

josephmyers:
This fixes a bug during sync where the exe couldn't be run due to permissions. I also removed FixFwData, since the production environment was working without it.

Both a change to FixFwData and Chorus are needed to launch successfully. Once the Chorus PR is merged, this can be readied.


megahirt:
c.f. Chorus PR https://github.com/sillsdev/chorus/pull/297

longrunningprocess (approved):
looks harmless enough for a troglodyte like me to approve!

PR #309, opened 2022-09-23
increased min model version from *68 to *72

josephmyers:
this removes support for FW8 projects. before, it was causing LfMerge to crash. now, doing this in LF displays the desired red box and maintains a good state for that user.


megahirt (approved):
YES LGTM

PR #325, opened 2023-04-27
Change environment variable names for Language Depot

rmunn:
All `LD_*` env vars in LfMerge will now be named `LFMERGE_LANGUAGE_DEPOT_*` instead. This is because Linux sudo (which LfMerge uses to run as the www-data user) is configured by default to discard any environment variables whose name starts with `LD_`, for security reasons. E.g., if you can mess with the `LD_LIBRARY_PATH` env var used by a program executing with privilege, you can inject arbitrary code into a privileged process and get it run.

So all environment variables used to tell LfMerge where to look for Language Depot now start with the `LFMERGE_LANGUAGE_DEPOT_` prefix. They are now documented in the README, and they can all be found in the MagicStrings.cs file in case I missed any in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,28 @@ The `master` branch is expected to be **fast-forwarded**, with no merge commits,
 
 For each DbVersion that LfMerge supports, we build a different lfmerge binary. We used to support DbVersions 7000068 through 7000070, which correspond to various versions of FieldWorks 8.x, but we now only support FieldWorks 9.x. The only DbVersion found in FieldWorks 9.x is currently (as of August 2022) 7000072. There is a script called `pbuild.sh` (for "parallel build") that will handle building all currently-supported DbVersions. It will run the build for each DbVersion in a Docker container, using a common Docker build image, and then copy the final results into a directory called `tarball`. Finally, it will run a Docker build that will take the files in the `tarball` directory and turn then into a Docker image for `lfmerge`. By default, this Docker image will be tagged `ghcr.io/sillsdev/lfmerge:latest`, while the GitHub Actions workflow will produce a specific version number tag as well as `latest`, i.e. the GHA workflow will tag `lfmerge:2.0.123` as well as `lfmerge:latest`.
 
+## Environment variables
+
+LfMerge can be configured by setting environment variables before launching LfMerge (probably in a Docker Compose or Kubernetes control file). These environment variables are:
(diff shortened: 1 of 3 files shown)


hahn-kev (approved):
Looks good, thanks Robin.

README.md
```diff
@@ -29,6 +29,28 @@ The `master` branch is expected to be **fast-forwarded**, with no merge commits,
...
+- **LFMERGE_MONGO_DB_NAME_PREFIX**: Default `sf_` (one trailing underscore), not `lf_`, again for historical reasons. The prefix added to a project's project code to create the name of the MongoDB database for that project's data. E.g. data for the `test-rmunn-05` project will be found in the `sf_test-rmunn-05` database.
+- **LFMERGE_VERBOSE_PROGRESS**: Boolean, default `false`. Whether to log verbose progress during Send/Receive. Recommended `true` in development & staging, `false` in production.
+- **LFMERGE_LANGUAGE_DEPOT_REPO_URI**: Default not set. The *complete* URI to a Language Depot project, including username & password. Not useful in production, highly useful in debugging and unit test scenarios.
+- **LFMERGE_LANGUAGE_DEPOT_HG_PUBLIC_HOSTNAME**: Default `hg-public.languagedepot.org`. The hostname of the Language Depot instance to use for **public** projects.
+- **LFMERGE_LANGUAGE_DEPOT_HG_PRIVATE_HOSTNAME**: Default is the value of the `PUBLIC_HOSTNAME` string with `public` replaced with `private`. The hostname of the Language Depot instance to use for **private** projects.
```
megahirt, line 51: @rmunn I have a random comment here.  My recollection is that LFMerge categorically does not deal with repositories on private.languagedepot.org and thus this variable is a relic that could be removed.  If this is not the case, how does a user actually connect up LD repo in private to Language Forge currently?  I don't know how that works.

rmunn: Without doing a deep dive into the code, my memory is that LfMerge will try public first, then if the repo doesn't exist in public it will then try the same repo name in private. (Which is behavior inherited from FLExBridge, and therefore the same behavior that FLEx uses). Now, because of how LF presents the list of projects to Send/Receive, there's only going to be public ones available to choose from, but I believe LfMerge is capable of doing S/R from both public and private because FLExBridge can do so.

PR #333, opened 2024-03-04
Update mercurial version

hahn-kev:
the current version does not work properly with https


rmunn:
Now the failure is "Cannot find libdl.so". https://stackoverflow.com/questions/75855053/how-to-address-crash-due-to-missing-libdl-so-on-ubuntu-22 seems relevant here.

rmunn:
Down to four test failures now, all of which have the same cause: the chorusmerge binary wants .NET 6 and is refusing to run under .NET 8. We just need to get chorusmerge built for both net6.0 and net8.0 and then, unless one of those four tests is concealing another bug, we should see all tests passing.

rmunn:
I've opened https://github.com/sillsdev/chorus/pull/332 to get a net8.0 build of chorusmerge.

rmunn:
Commit 528878f52cc4d63009aec821f8cb67474fcede5c updates the GHA workflow to just build with `dotnet build; dotnet test` instead of the convoluted process of creating a base builder image, a builder image, running the builder image, and creating a final image. Now there's just a single image (the final one) created. This makes it a lot simpler to verify the results.

rmunn:
> Down to four test failures now, all of which have the same cause: the chorusmerge binary wants .NET 6 and is refusing to run under .NET 8. We just need to get chorusmerge built for both net6.0 and net8.0 and then, unless one of those four tests is concealing another bug, we should see all tests passing.

ChorusMerge version 6.0.0-beta0044 has a net8.0 build in the NuGet package. Once I get back to this PR, I'll update the ChorusMerge dependency to 6.0.0-beta0044 and that should hopefully fix the last four failing tests.

rmunn:
Okay, LfMerge is building with .NET 8, Mercurial 6, and Python 3, and running the resulting Docker image doesn't immediately fall over. Now to test it in some S/R scenarios against local copies of Lexbox.

rmunn:
Just got the LDAPI working against local LexBox install. The trick was adding `- "admin.localhost:host-gateway"` to the `extra_hosts` definition for `lf-app` in the docker-compose.yml file; that makes admin.locahost resolve to the host network rather than the Docker Compose internal network, so it can actually hit the LexBox ingress.

LfMerge runs, but got an autorization error cloning the project. So I haven't yet been able to test the functionality of chorusmerge, which is the main point I want to test before merging this. So I'm not yet ready to merge it, but it's a step in the right direction.

rmunn:
S/R immediately fell over, several times, because we didn't have all of Mercurial copied into the final Docker image. Turns out between Mercurial 3 and 6, a lot of Mercurial files were moved around internally to different subdirectories, and our installation script didn't account for all of them. But that's sorted out now, and I've been able to prove that S/R, including running chorusmerge, works. Cloning, pulling, merging, pushing, all test out locally. Time to approve this and merge it.

Dockerfile.builder-base
```diff
@@ -1,13 +1,13 @@
 ARG DbVersion=7000072
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0 AS lfmerge-builder-base
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS lfmerge-builder-base
```
rmunn, line 3: Can probably get rid of the builder-base and runtime-base Dockerfiles now, but we'll leave that for a later PR.

pbuild.sh
```diff
@@ -37,47 +37,20 @@ FW9_BUILD_BRANCH="$(git name-rev --name-only HEAD)"
...
-docker build -t ghcr.io/sillsdev/lfmerge-base:sdk -f Dockerfile.builder-base .
-docker build -t ghcr.io/sillsdev/lfmerge-base:runtime -f Dockerfile.runtime-base .
-docker build --build-arg "BUILDER_UID=${CURRENT_UID}" -t lfmerge-builder-base --target lfmerge-builder-base .
-
-# Create the build images for each DbVersion
-for DbVersion in ${DBMODEL_VERSIONS[@]}; do
-	docker build --build-arg DbVersion=${DbVersion} --build-arg "BUILDER_UID=${CURRENT_UID}" -t lfmerge-build-${DbVersion} .
-done
+# Create prerequisite directories that LfMerge expects in the unit tests
+# TODO: Probably not needed now?
```
rmunn, line 44: TODO comment left in code.

rmunn (approved):
LGTM. Let's Get This Merged.

pbuild.sh
rmunn: Done in commit 7d4fabd3f7954212df681c1857ba817fdea24db3.

PR #337, opened 2024-06-24
Rename pbuild.sh script to build.sh

### LINKED ISSUE #297: Rename pbuild.sh to build.sh

rmunn:
Fixes #297.

In addition to the rename, I took this opportunity to edit the README.md file and simplify the build instructions a little.


README.md
```diff
@@ -53,11 +57,11 @@ Other settings you shouldn't need to touch:
...
+The image that `build.sh` produces is tagged with the same image tag as the one built by the official GitHub Actions workflow, with an `alpha` version number calculated from the name of the branch you're running and the number of commits since the branch was created. This means that if you're running Language Forge locally via the Makefile in Language Forge's `docker` directory, you should be able to simply edit the `docker/lfmerge/Dockerfile` in Language Forge. Make sure that file specifies the tag that was just built. Then run `make` and the `lfmerge` container will be re-created with your local build. Now run `docker logs -f lfmerge` in a console tab to watch the LfMerge logs, then do a Send/Receive via your `localhost` copy of Language Forge and check the resulting log files for any errors.
```
megahirt, line 64: re: alpha versioning

I would suggest that we no longer have a need for alpha versions.  We need simple versioning for LFMerge since we are both the single builder and consumer of the product.  If we had a mono-repo, I suspect this would not be versioned at all.  That said, if it's not broken there's no need to fix it.  Just keep in mind "reduce complexity" whenever you find yourself fighting with complexity

megahirt (approved):
LGTM thanks Robin!

README.md
rmunn: Sure. I'll fix that in a separate PR since that will involve changing the GHA workflow and a few other things. (Basically we'll be telling GitHub that `master` should become the default branch, and we'll get rid of the `develop` branch and start merging PRs directly into `master`).

PR #338, opened 2024-06-25
Don't serialize comments to JSON when calling LfMergeBridge

### LINKED ISSUE #329: Improve LanguageForgeGetChorusNotesActionHandler return value channel
right now LanguageForgeGetChorusNotesActionHandler returns a large amount of data via json serialization. By using a global static weak map we can return the data in memory instead of having to do a lot of work to pass the data back as a string.

The plan would be that there's a global static of type `ConditionalWeakTable<object, GetChorusNotesReponse>` which would use the passed in options Dictionary as the key and would create and would create the response (a new class), which would be retrieved in LfMerge `ConvertLcmToMongoComments.CallLfMergeBridge` using the same options Dictionary, we can remove the value from the global here, but this ensures that when options goes out of scope the response will get released from memory automatically.

rmunn:
Fixes #329.

Requires https://github.com/sillsdev/flexbridge/pull/401 and https://github.com/sillsdev/flexbridge/pull/402 to be built into a NuGet package before this can fully work. At the moment, those PRs have been merged into the flexbridge `develop` branch but the NuGet package build for FLExBridge is not yet working. So just for now, this only works if you check out the https://github.com/sillsdev/flexbridge repo to a sibling folder of LfMerge. (LfMerge currently uses a ProjectReference for flexbridge in order to be able to run the newly-merged code). Once a new NuGet package is built from flexbridge's `develop` branch, this PR can switch from using a ProjectReference for flexbridge to using a proper PackageReference, and then there won't be any need to check out the flexbridge repo alongside the lfmerge one.

Currently, PR checks are going to fail because Ci isn't set up to clone flexbridge alongside lfmerge, so the ProjectReference will fail. That's expected, and is one reason why https://github.com/sillsdev/flexbridge/pull/402 has to be merged first before this PR is merged.

## To test:

1. Make sure you have Language Forge and LexBox running side by side.
2. Check out `flexbridge` repo in a sibling folder to `LfMerge`.
3. In LfMerge repo, run `./pbuild.sh`. (Not yet renamed to `./build.sh` as of when this PR was branched off from `develop`).
4. Note the Docker image tag produced (probably something like `2.0.140-feat-less-serialization-of-comments.10`) and copy it to the clipboard.
5. In the web-languageforge repo, edit `docker/lfmerge/Dockerfile`
6. Change the lfmerge tag to the one from step 4.
7. In the web-languageforge repo, run `make`.
8. Run `docker logs -f lfmerge` so you'll see if there are any Send/Receive errors logged in the next steps.
9. Go to http://localhost/project/sena-3 (LexBox) and look at the project history of sena-3
10. Go to http://localhost:8080/ (Language Forge) in another tab and clone the sena-3 project.
11. Edit the sena-3 project in Language Forge and add a comment to an entry. Do a Send/Receive.
12. After the Send/Receive completes, go to http://localhost/project/sena-3 and verify that there's a new commit in sena-3.
13. Click on the new commit to be taken to the Mercurial repo for sena-3.
14. Click on `log` in the Mercurial repo for sena-3.
15. Click on the latest commit, which should be the one created by the Language Forge S/R.
16. Verify that it adds a message to the `.ChorusNotes` file, creating it if it didn't already exist.

That's a lot of steps, sorry. But this is just inherently complex; once a FlexBridge NuGet package has been built, you'll be able to skip steps 1 and 2, but the rest of the steps are necessary to test this feature.


src/LfMerge.Core/LfMerge.Core.csproj
```diff
@@ -37,13 +37,13 @@ See full changelog at https://github.com/sillsdev/LfMerge/blob/develop/CHANGELOG
     <PackageReference Include="Autofac" Version="4.*" />
     <PackageReference Include="INIFileParserDotNetCore.Signed" Version="2.5.2" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-    <PackageReference Include="MongoDB.Driver.signed" Version="2.14.*" />
-    <PackageReference Include="MongoDB.Driver.Core.signed" Version="2.14.*" />
+    <PackageReference Include="MongoDB.Driver.signed" Version="2.13.*" />
+    <PackageReference Include="MongoDB.Driver.Core.signed" Version="2.13.*" />
```
rmunn, line 41: This downgrade was necessary so that flexbridge could continue to target net461 and/or netstandard 2.0; starting with MongoDB driver 2.14, the .NET Framework target of the MongoDB packages was changed to .NET Framework 4.7.x and netstandard2.1 instead. The downgrade from 2.14 to 2.13 does not lose any bugfixes or new features that we relied on.

rmunn (commented):
This PR touches many files, but most of them are either deleting a file (since several of LfMerge's model classes have been moved to LfMergeBridge) or else they are a one-line change adding a single `using` line, so that the model classes can be found in their new namespace.

The only meaningful changes are to ConvertLcmToMongoComments.cs, the Mongo package downgrade in LfMerge.Core.csproj (see comment on that line), and a one-line change to MongoConnection.cs informing Mongo of the new namespace that should be searched for model classes. The rest of this PR is just deletions and/or adding a single `using` line.

megahirt (approved):
LGTM

src/LfMerge.Core/LfMerge.Core.csproj
megahirt: ok.  This is more evidence IMO that staying on .net461 forever is not a good idea.  Microsoft may be committed to supporting it on Windows going forward however dependency vendors (e.g. MongoDB) likely will not.

rmunn:
Just realized that there's more efficiency gains to be had: https://github.com/sillsdev/flexbridge/pull/402 does the same thing as https://github.com/sillsdev/flexbridge/pull/401 but in the other direction (https://github.com/sillsdev/flexbridge/pull/401 was handling comments from FW and giving them to LF, while https://github.com/sillsdev/flexbridge/pull/402 handles comments from LF and gives them to FW). So commit 673d2888 is the LfMerge side of https://github.com/sillsdev/flexbridge/pull/402.

@megahirt - If you've already reviewed the rest of this PR, you'll just need to look at commit 673d2888 to review the rest of the feature.

rmunn:
Approved and ready to merge, but we probably shouldn't *actually* merge this until we have automatically-built NuGet packages for FLExBridge again.

megahirt (approved):
still LGTM

Issue #339, opened 2024-06-25
Set up end-to-end testing of Send/Receive


rmunn:
Now that LexBox and Language Forge can be run side by side, it should be possible to set up end-to-end testing of Send/Receive scenarios with a real Language Depot deployment, rather than a simulated one. A typical test might go as follows:

1. Create a LexBox project, and upload a .zip file containing the initial state of the project.
2. Do an initial clone into Language Forge, verify correct number of entries and so on.
3. Use the [MkFwData tool](https://github.com/sillsdev/languageforge-lexbox/pull/735) to get a .fwdata file for that project.
4. Use liblcm to load that .fwdata file, then add a new entry or edit an existing entry.
5. Use the SplitFwData tool (that doesn't exist yet) to split the edited .fwdata file up into its component parts.
6. Create a new Mercurial commit with those changes.
7. Push that commit to the LexBox repo. (Steps 3-7 simulate making an edit in FieldWorks followed by a Send/Receive).
8. Use Mongo to make edits in the Language Forge project (faster), or use Playwright to drive the UI to make those changes (slower).
9. Have Language Forge trigger a Send/Receive, kicking off LfMerge.
10. Once LfMerge has finished the Send/Receive, verify that LF's Mongo database contains the changes from steps 2-7 (or the merge result of those changes).
11. In LexBox, verify that LfMerge created a new commit with the correct results.
12. Download an .fwdata file from LexBox and load it into liblcm.
13. Use liblcm to verify that the FieldWorks objects received the correct changes from LfMerge (updated field contents, comments, or whatever).

That's a lot of steps that currently have to be done by hand, but with LexBox and Language Forge running side by side on a developer machine (plus [MkFwData and SplitFwData tools](https://github.com/sillsdev/languageforge-lexbox/pull/735) becoming available in the flexbridge repo so that the E2E S/R tests can use them), all of that will finally be able to be automated.

rmunn:
More ideas: a helper class to manage various liblcm-related testing tasks:

* Given project code, clone it, create .fwdata file, open liblcm project
* Subset: given already-cloned Mercurial repo, create .fwdata file and open liblcm project
* Given liblcm project, close project, hg commit, push (maybe through LfMergeBridge for simplicity)

Also do various liblcm tasks:

* Create lexentry with given lexeme form
* Add sense, change part of speech
* Add example sentence to given sense
* Etc.

[...]

PR #342, opened 2024-07-25
Set up end to end Send/Receive testing with LexBox

rmunn:
Fixes https://github.com/sillsdev/LfMerge/issues/339.

This PR uses the LexBox API to set up end-to-end testing of Send/Receive scenarios.

**Work completed:**

* Reset project on LexBox to empty
* Upload new state of LexBox project
* Remove the last N commits from a LexBox project
* LcmTestHelper class full of static methods
* SRTestBase class that sets up and tears down temp folders appropriately, and can pass appropriate temp folders to LcmTestHelper methods
* Commit .fwdata changes and push result to LexBox
* First real end-to-end S/R test

**Future plans:**

In a new PR, I will port most of the SynchronizeActionTests tests and convert them to real end-to-end tests. Then I'll add some more tests testing things like round-tripping of comments, various merge conflict scenarios, and so on.


rmunn:
First end-to-end test is written and working. Doing self-review now, then will ask for code review.

rmunn:
The one failing test is the one this PR adds, with `HttpRequestException : Connection refused (localhost:80)`, from a previous `SocketException : Connection refused`. This is quite simply because the LfMerge GitHub workflow doesn't set up a copy of LexBox before running the tests. Before merging this PR, we might want to set up a local copy of LexBox, similarly to how https://github.com/sillsdev/languageforge-lexbox/pull/873 did it, so that the end-to-end tests can run.

Might want to split the test runs in two, actually; one runner will run the non-E2E tests with the current setup, and the other runner will set up a Kubernetes environment in GHA to run the E2E tests. That way the non-E2E tests can be run without paying the cost of setting up a k8s environment.

rmunn:
Okay, commit [779358e](https://github.com/sillsdev/LfMerge/commit/779358e507189c04e9c9d5c04cd890257a32dd60) demonstrates the race condition with hgweb's dir cache in local LexBox, and commit 80d45bba adjusts these tests to expect newly-created projects to sometimes not be available right away, avoiding the race condition. To test a potential fix in LexBox, use commit [779358e](https://github.com/sillsdev/LfMerge/commit/779358e507189c04e9c9d5c04cd890257a32dd60) and run `dotnet test --filter E2E_` repeatedly until you get test failures (shouldn't take long).

src/LfMerge.Core.Tests/E2E/TryOutE2ETests.cs
```diff
@@ -0,0 +1,61 @@
...
+		public async Task E2E_LFDataChangedLDDataChanged_LFWins()
+		{
+			// Setup
+
+			var lfProject = await CreateLfProjectFromSena3();
+			var fwProjectCode = Regex.Replace(lfProject.ProjectCode, "^sr-", "fw-");
+			var fwProject = CloneFromLexbox(lfProject.ProjectCode, fwProjectCode);
+
+			// Modify FW data first, then push to Lexbox
+			Guid entryId = Guid.Parse("0006f482-a078-4cef-9c5a-8bd35b53cf72");
```
hahn-kev, line 39: rather than hardcoding a guid I'd like to see is fetch an entry from the db.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -0,0 +1,86 @@
...
+			var transferLcmToMongo = new TransferLcmToMongoAction(TestEnv.Settings, SRTestEnvironment.NullLogger, _mongoConnection, _recordFactory);
+			transferLcmToMongo.Run(lfProject);
+
+			return lfProject;
+		}
+
+		public void SendReceiveToLexbox(LanguageForgeProject lfProject)
+		{
+			// ChorusHelperDouble.GetSyncUri assumes presence of a LanguageDepotMock, but here we want a real LexBox instance so we override it via environment variable
+			var saveEnv = Environment.GetEnvironmentVariable(MagicStrings.SettingsEnvVar_LanguageDepotRepoUri);
```
hahn-kev, line 53: uh, we can do better than this. We're passing in settings to the action below. Just modify LfMergeSettings so that it allows you to specify the repo Uri and then you don't need to mess around with environment variables.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -0,0 +1,203 @@
...
+using NUnit.Framework;
+using SIL.TestUtilities;
+using TusDotNetClient;
+
+namespace LfMerge.Core.Tests
+{
+	/// <summary>
+	/// Test environment for end-to-end testing, i.e. Send/Receive with a real LexBox instance
+	/// </summary>
+	public class SRTestEnvironment : TestEnvironment
```
hahn-kev, line 22: it looks like most of the code in this class is static, however it seems to be called from places that already have an instance of the environment. So I'd like that to be cleaned up. The environment stuff can stay static but everything else like the cookies and http clients etc should all be instance variables.

src/LfMerge.Core.Tests/SRTestBase.cs
```diff
@@ -0,0 +1,170 @@
...
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using SIL.TestUtilities;
+
+namespace LfMerge.Core.Tests
+{
+	/// <summary>
+	/// Test base class for end-to-end testing, i.e. Send/Receive with a real LexBox instance
+	/// </summary>
+	public class SRTestBase
```
hahn-kev, line 14: this class says it's for e2e testing, but then there's another class that's called e2eTestBase which inherits from this. Do we really need to use inheritance here? I would rather not use inheritance where we can, especially since it's not clear what order all the setup and teardown methods are called in.

src/LfMerge.Core.Tests/TestDoubles.cs
```diff
@@ -417,6 +417,8 @@ class ChorusHelperDouble: ChorusHelper
 	{
 		public override string GetSyncUri(ILfProject project)
 		{
+			var envOverride = Environment.GetEnvironmentVariable(MagicStrings.SettingsEnvVar_LanguageDepotRepoUri);
```
hahn-kev, line 420: this will break any test that is using the mock server, so if this variable is set during testing then those tests will fail

hahn-kev (changes requested):
This looks like a great start. I think there's still a bit of work to do before we can merge it in. Inparticualr I'd like to make a bunch of the static code to be non static instead. Also it's not clear to me why we have both SRTest base and e2eTestBase and why one inherits from the other. I'd like to either merge those into one class, or just make one encapsulate the other.

src/LfMerge.Core.Tests/TestDoubles.cs
rmunn: That's the desired behavior, and I considered it a bug that ChorusHelperDouble wasn't honoring that environment variable.

This variable is designed for this purpose. and [is usually unset](https://github.com/sillsdev/LfMerge/blob/f518d8d925ae9c6fa53c856df519d425f15d27c1/src/LfMerge.Core/Settings/DefaultLfMergeSettings.cs#L20); the ChorusHelper class looks at the `settings.LanguageDepotRepoUri` value (which comes from this environment variable) and treats it as an override, allowing one to run a single test with a different destination than normal.

rmunn: However, I did change this to pull from LfMergeSettings instead of the environment variable, so that the override can be set on a per-test basis without messing with env vars (as you mentioned in https://github.com/sillsdev/LfMerge/pull/342#discussion_r1722940309).

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: I made the HTTP client static so that we'd only have to log in once, and the GQL client static because the docs recommended only using one instance and I didn't want to mess with DI while building the prototype. But if we're just running a single E2E test anyway, there's no cost to making these instance variables, at which point many of the static methods will become non-static. (I changed them to static at the recommendation of the compiler's code analysis suggestions, but many of them access the Http client so they'll soon become non-static again).

I do think I'm going to keep one thing static: the Jwt property, which I'll rename to AdminJwt. Then the Login method will become LoginAsAdmin, and it will first check if the static Jwt property is set. If so, it will insert a cookie with that Jwt into the instance-local cookie container, allowing us to still get the "just log in once and reuse that for all tests" effect.

rmunn: Done in commit 3749cbe.

src/LfMerge.Core.Tests/TestDoubles.cs
rmunn: Done in commit bea1713.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
rmunn: Done in commit bea1713.

src/LfMerge.Core.Tests/E2E/TryOutE2ETests.cs
rmunn: Done in commit 946468d.

src/LfMerge.Core.Tests/SRTestBase.cs
rmunn: NUnit actually guarantees the order of setup and teardown methods across class inheritance; however, that's not necessarily going to hold true if we ever switch to a different test framework (e.g., XUnit doesn't use setup and teardown methods, preferring for you to put those in class constructors and Dispose() methods respectively).

But yes, I wrote this when I thought I was going to have a bunch more E2E tests, and this is over-engineered for the single test we ended up deciding on. So I'll combine SRTestBase and E2ETestBase into a single base class.

rmunn: Done in commit 5c7059b.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -6,21 +6,84 @@
...
-	public class E2ETestBase : SRTestBase
+	public class E2ETestBase
 	{
+		public LfMerge.Core.Logging.ILogger Logger => MainClass.Logger;
+		public TemporaryFolder TempFolderForClass { get; set; }
+		public TemporaryFolder TempFolderForTest { get; set; }
+		public TemporaryFolder TestDataFolder { get; set; }
+		public TemporaryFolder LcmDataFolder { get; set; }
+		public string Sena3ZipPath { get; set; }
+		private Guid? ProjectIdToDelete { get; set; }
```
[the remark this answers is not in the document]
hahn-kev: better yet these Ids come from the test environment which needs to be disposed. It should internally track created projects rather than doing that stuff here.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -28,43 +28,62 @@ public class SRTestEnvironment : TestEnvironment
...
-		private static bool AlreadyLoggedIn = false;
+		public static Cookie AdminLoginCookie { get; set; }
+		public readonly CookieContainer Cookies = new();
+		private HttpClientHandler Handler { get; init; }
+		private Lazy<HttpClient> LazyHttp { get; init; }
+		public HttpClient Http => LazyHttp.Value;
+		public readonly SIL.Progress.IProgress NullProgress = new SIL.Progress.NullProgress();
+		public readonly ILogger NullLogger = new NullLogger();
+		private bool AlreadyLoggedIn = false;
 		private TemporaryFolder TempFolder { get; init; }
```
hahn-kev, line 39: if I understand what these temp folders are for then this class should dispose of them once it's disposed of

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -6,21 +6,84 @@
...
+		}
+
+		[OneTimeTearDown]
+		public void FixtureTeardown()
+		{
+			Environment.SetEnvironmentVariable("FW_CommonAppData", null);
+			var result = TestContext.CurrentContext.Result;
+			var nonSuccess = result.FailCount + result.InconclusiveCount + result.WarningCount;
+			// Only delete class temp folder if we passed or skipped all tests
+			if (nonSuccess == 0) TempFolderForClass.Dispose();
```
hahn-kev, line 74: I think this means that because we ignore the tests when lexbox is not available that we don't cleanup the temp folder, is that right?

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -0,0 +1,238 @@
...
+
+		public async Task<string> CreateNewProjectFromTemplate(string origZipPath)
+		{
+			var randomGuid = Guid.NewGuid();
+			var testCode = $"sr-{randomGuid}";
+			var testPath = TestFolderForProject(testCode);
+			// Now create project in LexBox
+			var result = await TestEnv.CreateLexBoxProject(testCode, randomGuid);
+			Assert.That(result.Result, Is.EqualTo(LexboxGraphQLTypes.CreateProjectResult.Created));
+			Assert.That(result.Id, Is.EqualTo(randomGuid));
```
hahn-kev, line 179: you've got this assert duplicated above in the create empty project method. Would it make sense to move these asserts into the `TestEnv.CreateLexBoxProject` method?

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -0,0 +1,222 @@
...
+
+		public async Task UploadZip(string code, string zipPath)
+		{
+			var sourceUrl = new Uri(LexboxUrl, $"/api/project/upload-zip/{code}");
+			var file = new FileInfo(zipPath);
+			var client = new TusClient();
+			// client.AdditionalHeaders["Authorization"] = $"Bearer {Jwt}"; // Once we set up for LexBox OAuth, we'll use Bearer auth instead
+			var cookies = Cookies.GetCookies(LexboxUrl);
+			var authCookie = cookies[".LexBoxAuth"].ToString();
+			client.AdditionalHeaders["cookie"] = authCookie;
```
hahn-kev, line 172: I was really surprised that we can't just reuse our http client meaning we don't need to pass the cookies like this, so I looked into this TusClient and it looks like it hasn't been updated in 4 years and uses WebRequest under the covers, so I did a search and found this modern client https://www.nuget.org/packages/BirdMessenger which uses http client meaning we can just pass ours in and do away with copying cookies around.

hahn-kev (changes requested):
left some more feedback for this PR.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
rmunn: According to https://docs.nunit.org/articles/nunit/technical-notes/usage/Counting-Tests.html, ignored tests increment result.SkipCount, which I'm NOT including in the `nonSuccess` counter. So if lexbox is not available and those tests are ignored (but all other tests passed), then `nonSuccess` will be 0 and the temp folder will be cleaned up.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: This one is just set to the base class's `_languageForgeServerFolder` field, which is already disposed by the base class. I suppose I could just not have this property and use `_languageForgeServerFolder` everywhere, but it seemed nicer to have a better name for it.

So this one is already being disposed and doesn't need to be disposed again.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
rmunn: Done in commit 0af5fb2.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: Great, I'll use that. Projects that haven't been updated in more than a year or two always make me nervous, because what security vulnerability is going to go unpatched for years? Thanks for finding that more recently-updated project.

rmunn: OK, done in commit b3e258d.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -30,35 +93,124 @@ public void Setup()
 				throw new AssertionException("E2E tests need a mock MongoProjectRecordFactory in order to work.");
 		}
 
+		[TearDown]
+		public async Task TestTeardown()
```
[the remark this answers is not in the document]
rmunn: Done in commit 5fe1d6d.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -6,21 +6,84 @@
...
 	[Category("LongRunning")]
 	[Category("IntegrationTests")]
-	public class E2ETestBase : SRTestBase
+	public class E2ETestBase
 	{
+		public LfMerge.Core.Logging.ILogger Logger => MainClass.Logger;
+		public TemporaryFolder TempFolderForClass { get; set; }
+		public TemporaryFolder TempFolderForTest { get; set; }
+		public TemporaryFolder TestDataFolder { get; set; }
+		public TemporaryFolder LcmDataFolder { get; set; }
```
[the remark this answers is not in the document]
rmunn: Yep, they could be path strings. I used to use TestDataFolder elsewhere, but I seem to have ended up not needing that code so it got removed. LcmDataFolder was just for symmetry, but it too could be a path string.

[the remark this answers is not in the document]
rmunn: Done in commit 52017e9.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
```diff
@@ -30,35 +93,124 @@ public void Setup()
...
+						continue;
+					}
+					// If we got this far, no exception so we succeeded
+					success = true;
+				}
+			}
+			return dest;
+		}
+
+		public FwProject CloneFromLexbox(string code, string? newCode = null, TimeSpan? waitTime = null)
```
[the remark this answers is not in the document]
rmunn: Renamed to be more clear in commit 1eea903.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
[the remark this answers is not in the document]
rmunn: Made it a HashSet in commit 0d1e52f so that multiple projects can be deleted if necessary. I'd rather track this in the test class rather than the environment, because the decision of whether to delete the test IDs or not depends on whether the test succeeded.

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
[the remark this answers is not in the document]
hahn-kev: ok so can these properties be removed now?

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -0,0 +1,240 @@
...
+			var metadata = new MetadataCollection { { "filetype", mimeType } };
+			var createOpts = new TusCreateRequestOption {
+				Endpoint = tusEndpoint,
+				UploadLength = file.Length,
+				Metadata = metadata,
+			};
+			var createResponse = await Http.TusCreateAsync(createOpts);
+
+			// That doesn't actually upload the file; TusPatchAsync does the actual upload
+			var fileStream = new FileStream(file.FullName, FileMode.Open, FileAccess.Read);
```
hahn-kev, line 181: I think this stream needs to be disposed. It doesn't look like tus does that for you.

hahn-kev (changes requested):
left 2 requests. Otherwise it's looking almost done

src/LfMerge.Core.Tests/E2E/E2ETestBase.cs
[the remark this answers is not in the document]
rmunn: Whoops, didn't actually remove the properties, only their usage. Removed in commit 68440129.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: Done in commit 0f10f6b6.

hahn-kev (approved):
Looks good, we will enable these to run on CI in another PR

PR #348, opened 2024-08-23
Use real MongoDB connection for E2E tests

### LINKED ISSUE #345: LfMerge integration tests with real MongoDB container
LfMerge integration tests in #342 currently run against a mock Mongo connection. But for real end-to-end testing, we probably want to set up a real MongoDB instance and populate the test project in Mongo, then run the tests that way. We're not talking about setting up an entire local Language Forge, just a Mongo database.

This will be a good follow-up issue once #342 is merged.

rmunn:
Fixes #345.

The end-to-end LfMerge tests now use a real MongoDB container (spun up and torn down for each test, with no data persistence) rather than a mock Mongo connection. This does require the presence of Docker on the machine running the E2E tests, so that `docker run mongo:6` will succeed.

### DIFF
--- a/src/LfMerge.Core.Tests/SRTestEnvironment.cs
+++ b/src/LfMerge.Core.Tests/SRTestEnvironment.cs
@@ -96,7 +96,14 @@ namespace LfMerge.Core.Tests
 		{
 			if (_disposed) return;
 			if (disposing) {
-				StopMongo();
+				// Set the LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE env var to a non-empty value if you want Mongo container to NOT be torn down on test failure
+				if (CleanUpTestData || !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(MagicStrings.EnvVar_E2E_LeaveMongoContainerRunningOnFailure))) {
+					StopMongo();
+				} else {
+					Console.WriteLine($"Leaving Mongo container {MongoContainerId} around to examine data on failed test.");
+					Console.WriteLine($"It is listening on {Settings.MongoDbHostNameAndPort}");
+					Console.WriteLine($"To delete it, run `docker stop {MongoContainerId} ; docker rm {MongoContainerId}`.");
+				}
 			}
 			base.Dispose(disposing);
 		}
--- a/src/LfMerge.Core.Tests/SRTestEnvironment.cs
+++ b/src/LfMerge.Core.Tests/SRTestEnvironment.cs
@@ -92,12 +92,19 @@ namespace LfMerge.Core.Tests
 			}
 		}
 
+		private bool ShouldStopMongoOnFailure()
+		{
+			// Mongo container will be torn down on test failure unless LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE
+			// is set to a non-empty value (except "false" or "0", which mean the same as leaving it empty)
+			var envVar = Environment.GetEnvironmentVariable(MagicStrings.EnvVar_E2E_LeaveMongoContainerRunningOnFailure)?.Trim();
+			return string.IsNullOrEmpty(envVar) || envVar == "false" || envVar == "0";
+		}
+
 		protected override void Dispose(bool disposing)
 		{
 			if (_disposed) return;
 			if (disposing) {
-				// Set the LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE env var to a non-empty value if you want Mongo container to NOT be torn down on test failure
-				if (CleanUpTestData || !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(MagicStrings.EnvVar_E2E_LeaveMongoContainerRunningOnFailure))) {
+				if (CleanUpTestData || ShouldStopMongoOnFailure()) {
 					StopMongo();
 				} else {
 					Console.WriteLine($"Leaving Mongo container {MongoContainerId} around to examine data on failed test.");
--- a/src/LfMerge.Core.Tests/SRTestEnvironment.cs
+++ b/src/LfMerge.Core.Tests/SRTestEnvironment.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Json;
 using System.Threading.Tasks;
+using Autofac;
 using BirdMessenger;
 using BirdMessenger.Collections;
 using GraphQL;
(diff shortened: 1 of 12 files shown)


src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -51,6 +55,58 @@ public SRTestEnvironment(TemporaryFolder? tempFolder = null)
 			Settings.CommitWhenDone = true; // For SR tests specifically, we *do* want changes to .fwdata files to be persisted
 		}
 
+		override protected void RegisterMongoConnection(ContainerBuilder builder)
+		{
+			// E2E tests want a real Mogno connection
+			builder.RegisterType<MongoConnection>().As<IMongoConnection>().SingleInstance();
+		}
+
+		public void LaunchMongo()
```
hahn-kev, line 64: it looks like this method will mostly fail silently making tests fail in unexpected ways. This method should throw if it can't launch mongo

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -51,6 +55,58 @@ public SRTestEnvironment(TemporaryFolder? tempFolder = null)
...
+
+		protected override void Dispose(bool disposing)
+		{
+			if (_disposed) return;
+			if (disposing) {
+				if (CleanUpTestData) {
+					StopMongo();
+				} else {
+					if (MongoContainerId is not null) {
+						Console.WriteLine($"Leaving Mongo container {MongoContainerId} around to examine data on failed test.");
```
hahn-kev, line 101: I don't really like the idea of leaving the container around as it means if there's a bug and you're trying to fix a test and you run it repeatedly you're going to end up with a bunch of containers which you must stop yourself. I'd rather have a flag that we can manually enable to keep the container around, otherwise it should always be killed.

src/LfMerge.Core/Settings/LfMergeSettings.cs
```diff
@@ -74,10 +72,10 @@ public string LanguageDepotRepoUri {
 
 		public bool CommitWhenDone { get; internal set; }
 
-		public string MongoDbHostNameAndPort { get { return String.Format("{0}:{1}", MongoHostname, MongoPort.ToString()); } }
+		public string MongoDbHostNameAndPort { get { return String.Format("{0}:{1}", MongoHostname, MongoPort); } }
```
hahn-kev, line 75:
```suggestion
		public string MongoDbHostNameAndPort => $"{MongoHostname}:{MongoPort}";
```

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -51,6 +55,58 @@ public SRTestEnvironment(TemporaryFolder? tempFolder = null)
...
+				var result = CommandLineRunner.Run("docker", "run -p 27017 -d mongo:6", ".", 30, NullProgress);
+				MongoContainerId = result.StandardOutput?.TrimEnd();
+				if (MongoContainerId != null)
+				{
+					result = CommandLineRunner.Run("docker", $"port {MongoContainerId} 27017", ".", 30, NullProgress);
+					var hostAndPort = result.StandardOutput?.TrimEnd();
+					var parts = hostAndPort.Contains(':') ? hostAndPort.Split(':') : null;
+					if (parts is not null && parts.Length == 2) {
+						Settings.MongoHostname = parts[0].Replace("0.0.0.0", "localhost");
+						Settings.MongoPort = parts[1];
```
hahn-kev, line 77: should we attempt to connect to the db to ensure it's up before we return from this method?

hahn-kev (changes requested):
the main change I would like is to always cleanup the mongo container, and to throw if mongo fails to start

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: Done in commit 48f4608.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: Done in commit 48f4608, except for the flag.

src/LfMerge.Core/Settings/LfMergeSettings.cs
rmunn: Done in commit 48f4608.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: I'd rather not. The MongoClient class from the Mongo C# drivers already has a timeout of 40 seconds to connect to Mongo, and before any Mongo connection is attempted, there is going to be several seconds' worth of setup in Lexbox: downloading a backup zip of sena-3, creating a new project, resetting the project and uploading the backup zip... The time that takes is sufficient so that I've never had an issue with Mongo not being up by the time the test needs it, and even if Mongo isn't quite finished going up yet, the MongoClient will wait 40 seconds already which is enough to get Mongo up. I wouldn't expect any test failures from Mongo not being ready in time.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: Flag implemented in commit 563a0a2.

src/LfMerge.Core.Tests/SRTestEnvironment.cs
```diff
@@ -51,6 +55,59 @@ public SRTestEnvironment(TemporaryFolder? tempFolder = null)
...
+				MongoContainerId = null;
+			}
+		}
+
+		protected override void Dispose(bool disposing)
+		{
+			if (_disposed) return;
+			if (disposing) {
+				// Set the LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE env var to a non-empty value if you want Mongo container to NOT be torn down on test failure
+				if (CleanUpTestData || !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(MagicStrings.EnvVar_E2E_LeaveMongoContainerRunningOnFailure))) {
```
hahn-kev, line 100: this would mean that `LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE =false` would be considered enabled

hahn-kev (changes requested):
looks good, I'd like to allow setting `LFMERGE_E2E_LEAVE_MONGO_CONTAINER_RUNNING_ON_FAILURE` to false should disable it. Otherwise it's good to go

src/LfMerge.Core.Tests/SRTestEnvironment.cs
rmunn: In commit 95e29ea I allow a value of "false" or "0" to also be counted as disabled. (0 because a lot of Microsoft's stuff suggests setting DOTNET_ENV_VAR to 1 to enable it, so 0 is a natural "disabled" meaning). I stopped short of treating "f" and "n" and "no" as false values, though that would of course be quite easy to add given the refactoring I did in 95e29ea.

hahn-kev (approved):

PR #358, opened 2025-06-16
Increase Chorus merge timeout to 60 minutes

rmunn:
At least one project took 23 minutes 3 seconds to do an hg merge operation, so Chorus's default 15 minute timeout is too low.

Fixes #357.

PR #359, opened 2026-06-02
Raise default Chorus local-op timeout so large projects can Send/Receive

### LINKED ISSUE #357: Increase Chorus timeout by setting env var
There is one project on Language Forge that has had support requests several times for merge problems. At least once, the only problem was that creating the merge commit took too long. Chorus has a default timeout of 15 minutes for local operations (`hg commit`, `hg merge`, and so on) and on the most recent Send/Receive of that project (2025-06-16), the merge commit took 23 minutes and 3 seconds to complete. It failed the first time, but I was able to make it succeed by setting the CHORUS_LOCAL_TIMEOUT_SECONDS variable in `/usr/share/lfmerge/7000072/environ`. That edit was live on the container, however, and will be overwritten the next time the `lfmerge` container is deployed.

We should edit the `environ` file in this repo and set CHORUS_LOCAL_TIMEOUT_SECONDS to a higher number than 15 minutes. I suggest 60 minutes, 3600 seconds, as a good compromise between allowing large projects to continue merging while not allowing a runaway merge process to lock up the server forever. That way the next time lfmerge is deployed, this large project will continue to Send/Receive successfully.

myieye:
Chorus's default 15-minute timeout on local hg operations (notably the post-merge `hg commit`) is too short for very large projects: the commit runs past 900s and the project lands on hold with a `TimeoutException`. This recurs for at least one production project (#357), where a maintainer has had to hand-edit `environ` on the live container — which is wiped on every redeploy.

This bakes the fix into the image:
- `environ`: default `CHORUS_LOCAL_TIMEOUT_SECONDS` to 1800 (30 min) via `${VAR:-1800}` — a conservative default that a deployment can raise as needed.
- `lfmergeqm-background.sh` / `lfmergeqm-looping.sh`: add `CHORUS_LOCAL_TIMEOUT_SECONDS` to the `sudo --preserve-env` list, so a deployment-supplied value survives the privilege drop to `www-data` and overrides the default. Without this, an env var set on the container is stripped by sudo and never reaches Chorus.

Closes #357.

### DIFF
--- a/environ
+++ b/environ
@@ -9,3 +9,9 @@ export PATH="${BASE}/output/${BUILD}/net8.0:${BASE}/output/Mercurial:${PATH}"
 # set HGRCPATH so that we ignore ~/.hgrc files which might have content that is
 # incompatible with our version of Mercurial
 export HGRCPATH=/etc/mercurial/hgrc
+
+# Chorus's 15-min default timeout on local hg ops is too short for very large
+# projects, whose merge commits can exceed it and land the project on hold
+# (https://github.com/sillsdev/LfMerge/issues/357). Overridable per deployment via
+# CHORUS_LOCAL_TIMEOUT_SECONDS, passed through by the launch scripts' --preserve-env.
+export CHORUS_LOCAL_TIMEOUT_SECONDS=${CHORUS_LOCAL_TIMEOUT_SECONDS:-1800}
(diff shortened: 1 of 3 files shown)


myieye:
Closing in favor of #358, which already addresses #357 by setting CHORUS_LOCAL_TIMEOUT_SECONDS in environ. Thanks @rmunn.

